PR #70, opened 2021-08-04
Parse inline visibilities

Morriar:
Make it possible to parse things like this:

```rb
private def m1; end
protected def self.m2; end
private attr_reader :a
```

### DIFF
--- a/test/rbi/parser_test.rb
+++ b/test/rbi/parser_test.rb
@@ -658,5 +658,23 @@ module RBI
         ); end
       RBI
     end
+
+    def test_parse_errors
+      e = assert_raises(RBI::ParseError) do
+        RBI::Parser.parse_string(<<~RBI)
+          def bar
+        RBI
+      end
+      assert_equal("unexpected token $end", e.message)
+      assert_equal("-:2:0-2:0", e.location.to_s)
+
+      e = assert_raises(RBI::ParseError) do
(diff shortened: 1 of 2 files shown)


test/rbi/parser_test.rb
```diff
@@ -90,6 +90,27 @@ def m4; end
       assert_equal(rbi, out.string)
     end
 
+    def test_parse_methods_with_visibility
+      rbi = <<~RBI
+        private def m1; end
+        protected def self.m2; end
+        private attr_reader :a
+        private class Foo; end
```
Morriar, line 98: This is valid syntax, it will fail at runtime with a TypeError so we need to be able to parse it without crashing.

I wonder if we should raise a ParseError on our end or keep generating the weird tree as follow:

```rb
private
class Foo; end
```

In theory we should never find this in RBI's since it's also an error caught by Sorbet: https://sorbet.run/#%23%20typed%3A%20true%0A%0Aprivate%20class%20Foo%3B%20end

vinistock (approved):

test/rbi/parser_test.rb
vinistock: I vote for raising, since Ruby itself will eventually raise as well.

KaanOzkan (approved):

PR #71, opened 2021-08-04
Introduce AddTodoSigs rewriter

Morriar:
This rewriter adds empty signatures to each method/attribute that does not have one already:

```rb
  attr_reader :a
  def foo(x, &block); end
RBI
 
rbi.add_todo_sigs!(with_todo_comment: true)

assert_equal(<<~RBI, rbi.string)
  # TODO: fill in signature
  sig { returns(T.untyped) }
  attr_reader :a

  # TODO: fill in signature
  sig { params(x: T.untyped, &block: T.untyped).returns(T.untyped) }
  def foo(x, &block); end
RBI
```


vinistock (approved):
Nice. Is the idea that adding this will lead to the right behavior of devs filling in the missing sigs?

KaanOzkan (approved):
Nitpick: It might be better to have a more descriptive TODO comment. Such as `fill in signature with appropriate type information`

PR #72, opened 2021-08-04
Create, print, parse and handle RBI::Struct nodes

Morriar:
As required to implement https://github.com/Shopify/tapioca/pull/247.


test/rbi/rewriters/sort_nodes_test.rb
```diff
@@ -51,6 +51,21 @@ class C; end
         RBI
       end
 
+      it("sorts strructs") do
```
paracycle, line 54:
```suggestion
      it("sorts structs") do
```

paracycle (approved):
Read through it all and all I found was a type in a test name 😂 Great work

vinistock (approved):

PR #75, opened 2021-08-05
Add Annotate & Deannotate rewriters

Morriar:
Pretty simple rewriters to add and remove @annotations from RBI nodes.

Annotate nodes:

```rb
  module A
    attr_reader :a
    def m; end
  end
RBI

rbi.annotate!("test")

assert_equal(<<~RBI, rbi.string)
  # @test
  module A
    # @test
    attr_reader :a

    # @test
    def m; end
  end
RBI
```

Deannotate nodes:

```rb
  # @test
  module A
    # @test
    attr_reader :a

    # @test
    def m; end
  end
RBI

rbi.deannotate!("test")

assert_equal(<<~RBI, rbi.string)
  module A
    attr_reader :a
    def m; end
  end
RBI
```

This can then be used to tag RBI definitions coming from a "trusted" source such as `sorbet/rbi/shims` or the `rbi/` and automatically resolve conflicts during merge operations.


lib/rbi/model.rb
```diff
@@ -80,6 +80,20 @@ def initialize(loc: nil, comments: [])
...
       @comments = comments
     end
+
+    sig { returns(T::Array[String]) }
+    def annotations
+      comments.select do |comment|
+        comment.text.start_with?("@")
+      end.map do |comment|
+        T.must(comment.text[1..])
+      end
```
vinistock, line 90: Suggestion: I think it's more readable like this
```ruby
comments
  .select { |comment| comment.text.start_with?("@") }
  .map { |comment| T.must(comment.text[1..]) }
```

test/rbi/rewriters/annotate_test.rb
```diff
@@ -0,0 +1,114 @@
...
+          end
+        end
+
+        class C < T::Struct
+          const :a, String
+          prop :b, String
+        end
+      RBI
+
+      rbi.annotate!("test")
```
vinistock, line 27: Here, invoking annotate adds the comment annotation to all members of the RBI. I find that a bit unexpected, I would have thought it would only annotate the root level of the RBI and that we'd need to invoke `annotate!` on the specific nodes.

Is there a reason for this design?

lib/rbi/model.rb
```diff
@@ -80,6 +80,20 @@ def initialize(loc: nil, comments: [])
...
+    end
+
+    sig { params(annotation: String).returns(T::Boolean) }
+    def has_annotation?(annotation)
+      annotations.one? { |a| a == annotation }
```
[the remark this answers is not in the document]
Morriar: TIL, very nice! Indeed, I think we can remove the method :+1:

test/rbi/rewriters/annotate_test.rb
Morriar: Hmmm good point.

The idea is to provide a semi-automatic way to resolve merge conflict against manually create definitions inside shims.

Given this Tapioca generated RBI:
```rb
# left.rbi

class A
  def a1; end
  def a2(x); end
  def a3(x); end
end
```

And this shim:

```rb
# right.rbi

class A
  def a1; end

  def a2(y); end

  @shim
  def a3(x, y); end
end
```

Running `rbi merge --keep-annotated=shim` should give you this:

```rb
# merged.rbi

class A
  def a1; end

  <<<<<<< left.rbi
  def a2(x); end
  =======
  def a2(y); end
  >>>>>>> right.rbi

  @shim
  def a3(x, y); end
end
```

The current PR aims at providing an automated way to create those `@shim` annotations.

Maybe the annotate rewriter should take some options like `annotate_scopes`, `annotate_properties`, `annotate_top_level_only` to control the behaviour? wdyt?

PR #76, opened 2021-08-05
Allow automatic conflict resolution on merges

Morriar:
Add options to the merge rewriter so conflicts can be automatically resolved:

No automatic resolution:

```rb
left = RBI::Parser.parse_string(<<~RBI)
  module Foo
    A = 10

    class Bar
      def m1; end

      sig { void }
      def m2; end

      def m3; end
    end
  end
RBI

right = RBI::Parser.parse_string(<<~RBI)
  module Foo
    A = 42

    module Bar
      def m1(x); end

      sig { returns(Integer) }
      def m2; end

      def m4; end
    end
  end
RBI

res = Rewriters::Merge.merge_trees(left, right)

assert_equal(<<~RBI, res.string)
  module Foo
    <<<<<<< left
    A = 10
    =======
    A = 42
    >>>>>>> right

    <<<<<<< left
    class Bar
    =======
    module Bar
    >>>>>>> right
      <<<<<<< left
      def m1; end
      =======
      def m1(x); end
      >>>>>>> right
      <<<<<<< left
      sig { void }
      def m2; end
      =======
      sig { returns(Integer) }
      def m2; end
      >>>>>>> right
      def m3; end
      def m4; end
    end
  end
RBI
```

Resolve on keeping `left` definitions:

```rb
left = RBI::Parser.parse_string(<<~RBI)
  module Foo
    A = 10

    class Bar
      def m1; end

      sig { void }
      def m2; end

      def m3; end
    end
  end
RBI

right = RBI::Parser.parse_string(<<~RBI)
  module Foo
    A = 42

    module Bar
      def m1(x); end

      sig { returns(Integer) }
      def m2; end

      def m4; end
    end
  end
RBI

res = Rewriters::Merge.merge_trees(left, right, keep: Rewriters::Merge::Keep::LEFT)

assert_equal(<<~RBI, res.string)
  module Foo
    A = 10

    class Bar
      def m1; end

      sig { void }
      def m2; end

      def m3; end
      def m4; end
    end
  end
RBI
```

Resolve on keeping `right` definitions:

```rb
left = RBI::Parser.parse_string(<<~RBI)
  module Foo
    A = 10

    class Bar
      def m1; end

      sig { void }
      def m2; end

      def m3; end
    end
  end
RBI

right = RBI::Parser.parse_string(<<~RBI)
  module Foo
    A = 42

    module Bar
      def m1(x); end

      sig { returns(Integer) }
      def m2; end

      def m4; end
    end
  end
RBI

res = Rewriters::Merge.merge_trees(left, right, keep: Rewriters::Merge::Keep::RIGHT)

assert_equal(<<~RBI, res.string)
  module Foo
    A = 42

    module Bar
      def m1(x); end

      sig { returns(Integer) }
      def m2; end

      def m3; end
      def m4; end
    end
  end
RBI
```

### DIFF
--- a/lib/rbi/rewriters/merge_trees.rb
+++ b/lib/rbi/rewriters/merge_trees.rb
@@ -39,9 +39,17 @@ module RBI
     class Merge
       extend T::Sig
 
-      sig { params(left: Tree, right: Tree, left_name: String, right_name: String).returns(Tree) }
-      def self.merge_trees(left, right, left_name: "left", right_name: "right")
-        rewriter = Rewriters::Merge.new(left_name: left_name, right_name: right_name)
+      class Keep < ::T::Enum
+        enums do
+          NONE = new
+          LEFT = new
+          RIGHT = new
+        end
+      end
+
+      sig { params(left: Tree, right: Tree, left_name: String, right_name: String, keep: Keep).returns(Tree) }
+      def self.merge_trees(left, right, left_name: "left", right_name: "right", keep: Keep::NONE)
+        rewriter = Rewriters::Merge.new(left_name: left_name, right_name: right_name, keep: keep)
         rewriter.merge(left)
         rewriter.merge(right)
         tree = rewriter.tree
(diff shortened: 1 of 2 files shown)


vinistock:
Do we also need a keep both option?

PR #81, opened 2021-09-28
Properly print comments with new lines

### LINKED ISSUE #80: `RBI::Comment` should properly treat new lines
Currently I get:
```ruby
$ ruby -rrbi -e 'puts RBI::Comment.new("Foo\nBar").string'
# Foo
Bar
```
when I would have expected the output to be:
```ruby
# Foo
# Bar
```

Morriar:
Closes #80.

cc. @paracycle


paracycle (approved):
👌

PR #82, opened 2021-09-28
Silence warnings around `Unparser` loading

paracycle:
Currently, when the RBI gem is loaded, we get a warning message output in the console about a Ruby version mismatch, if the [teeny version](https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/) of the running Ruby doesn't match the teeny version of the Parser::Current Ruby. As far as the RBI gem is concerned, this warning message is hardly actionable, so it is better to suppress it than to display it to downstream clients.

We get the warning message very early since when we require `unparser` it requires `parser/current` immediately. However, the program might not even be performing any parsing operations during its execution, so this can easily be delayed to parsing time, by doing the require delayed.

Once we've delayed the require, we can silence the `Warning.warn` method while we are doing the first `unparser` require. Thus, we can completely eliminate the warnings raised by `Parser::Current`.


lib/rbi/parser.rb
```diff
@@ -51,6 +51,7 @@ def parse_file(path)
 
     sig { params(content: String, file: String).returns(Tree) }
     def parse(content, file:)
+      require_unparser
```
Morriar, line 54: What do you think about moving this to the `initialize`?

Morriar (approved):

PR #92, opened 2021-10-18
Change to released tapioca dependency

mbj:
* The previous tapioca release referenced, via Gemfile.lock was 7 month old.
* It referenced the since deleted `master` branch.
* `bundle update` or similar operations would fail as the `master`
  refname cannot be resolved anymore.
* Changing to a `main` git reference alternative was dismissed as it
  looks like `tapioca` is more stable meanwhile and we should not
  reference a git repository unless we have a good reason.


mbj:
Hehe, its actually the other way round, tapioca should use `rbi` so the git dependency on tapioca was just an artifact its not used at all. Closing this and opening a remove PR.

PR #100, opened 2021-11-22
Introduce `RemoveKnownDefinitions` rewriter

Morriar:
Remove all definitions existing in the index from the current tree

Let's create an `Index` from two different `Tree`s:
~~~rb
tree1 = Parse.parse_string(<<~RBI)
  class Foo
    def foo; end
  end
RBI

tree2 = Parse.parse_string(<<~RBI)
  FOO = 10
RBI

index = Index.index(tree1, tree2)
~~~

We can use `RemoveKnownDefinitions ` to remove the definitions found in the `index` from the `Tree` to clean:
~~~rb
tree_to_clean = Parser.parse_string(<<~RBI)
  class Foo
    def foo; end
    def bar; end
  end
  FOO = 10
  BAR = 42
RBI

cleaned_tree, operations = RemoveKnownDefinitions.clean(tree_to_clean)

assert_equal(<<~RBI, cleaned_tree)
  class Foo
    def bar; end
  end
  BAR = 42
RBI

assert_equal(<<~OPERATIONS, operations.join("\n"))
  Deleted ::Foo#foo at -:2:2-2-16 (duplicate from -:2:2-2:16)
  Deleted ::FOO at -:5:0-5:8 (duplicate from -:1:0-1:8)
OPERATIONS
~~~


lib/rbi/rewriters/clean_tree.rb
```diff
@@ -0,0 +1,125 @@
...
+    # tree_to_clean = Parser.parse_string(<<~RBI)
+    #   class Foo
+    #     def foo; end
+    #     def bar; end
+    #   end
+    #   FOO = 10
+    #   BAR = 42
+    # RBI
+    #
+    # cleaned_tree, operations = CleanTree.clean(tree_to_clean)
```
KaanOzkan, line 34: Missing an `index` parameter.

lib/rbi/rewriters/clean_tree.rb
```diff
@@ -0,0 +1,125 @@
...
+    #   end
+    #   BAR = 42
+    # RBI
+    #
+    # assert_equal(<<~OPERATIONS, operations.join("\n"))
+    #   Deleted ::Foo#foo at -:2:2-2-16 (duplicate from -:2:2-2:16)
+    #   Deleted ::FOO at -:5:0-5:8 (duplicate from -:1:0-1:8)
+    # OPERATIONS
+    # ~~~
+    class CleanTree < Visitor
```
KaanOzkan, line 48: I don't love this name since "cleaning" isn't associated with "remove from index" in my mind. But I don't have a good alternative 😕. I had to read the docs in detail for it to make sense.

lib/rbi/rewriters/clean_tree.rb
```diff
@@ -0,0 +1,125 @@
...
+        when RBI::Indexable
+          previous = previous_definition_for?(node)
+          delete_node(node, previous) if previous
+        end
+      end
+
+      private
+
+      sig { params(node: RBI::Indexable).returns(T.nilable(RBI::Node)) }
+      def previous_definition_for?(node)
```
KaanOzkan, line 98: Should the method not have a `?` in its name since it's not returning a boolean? The return value is used and not just as a truthy/falsy value.

lib/rbi/rewriters/clean_tree.rb
Morriar: Ok. I went with `RemoveKnownDefinitions`. Let me know what you think.

KaanOzkan (approved):
I think it's much more intuitive now, thank you.

We could also rename `#clean` method to `#remove` to be consistent if you want.

Morriar:
> We could also rename `#clean` method to `#remove` to be consistent if you want.

Done 👍

KaanOzkan (approved):
Tests methods still contain `clean_tree` in their names.

RyanBrushett (approved):
Pending Kaan's observation about test names.

PR #102, opened 2021-11-23
Fix IndexTest test names

Morriar:
Those tests were never running 🤦


RyanBrushett (approved):
Gemfile.lock commit is in https://github.com/Shopify/rbi/pull/101

Good catch on the tests ❤️

KaanOzkan (approved):

PR #106, opened 2021-11-24
Fix indeterministic sort order for anonymous nodes

Morriar:
Two things here:

1. When a node didn't have a name (for example `class << self; end`) the comparison returned 0 (without preserving the original order like intended) which might result in indeterministic order
2. Always group and sort `class << self; end` nodes before other scopes and constants

See tests for more examples.

This should fix unstable order for RBI generation that was happening with `tapioca dsl --verify`.

### DIFF
--- a/lib/rbi/rewriters/group_nodes.rb
+++ b/lib/rbi/rewriters/group_nodes.rb
@@ -67,6 +67,8 @@ module RBI
         else
           Group::Kind::Methods
         end
+      when SingletonClass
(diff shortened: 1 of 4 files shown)


KaanOzkan (approved):
🎉

mutecipher (approved):
👏🏻

RyanBrushett (approved):

ajkerr:
🙌

PR #108, opened 2021-12-01
Add flatten_scopes rewriter

jeffcarbs:
Flattens scopes, so:
```
module A
  class B
    FOO = 42

    class C
      module D; end
    end
  end
end
module E; end
```

becomes:
```
module A; end
class A::B
  FOO = 42
end
class A::B::C; end
module A::B::C::D; end
module E; end
```


lib/rbi/rewriters/flatten_scopes.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        case node
+        when Tree
+          visit_all(node.nodes)
+
+          parent_tree = node.parent_tree
+          return unless parent_tree
+
+          node.nodes.dup.each do |child|
+            next unless child.is_a?(Class) || child.is_a?(Module)
```
Morriar, line 21: Why do we need to iterate over the child nodes of the tree?

Can't we have a `when` for those instead?

Something like this:

```rb
case node
when Class, Module
  visit_all(node.nodes)

  parent_tree = node.parent_tree
  return unless parent_tree

  parent_scope = node.parent_scope
  return unless parent_scope.is_a?(Class) || parent_scope.is_a?(Module)

  node.detach
  node.name = "#{parent_scope.name}::#{node.name}"

  parent_tree << node
when Tree
  visit_all(node.nodes)
end
```

test/rbi/rewriters/flatten_scopes_test.rb
```diff
@@ -0,0 +1,102 @@
...
+
+        class A::B::C
+          C1 = 42
+        end
+
+        module A::B::C::D; end
+      RBI
+    end
+
+    def test_flatten_scopes_with_singleton_classes
```
Morriar, line 69: Let's also add tests with `T::Enum` and `T::Struct` please

test/rbi/rewriters/flatten_scopes_test.rb
```diff
@@ -0,0 +1,102 @@
...
+      scope1 = RBI::Module.new("A")
+      scope2 = RBI::Class.new("B")
+      scope3 = RBI::Class.new("C")
+      scope4 = RBI::Module.new("D")
+      scope5 = RBI::Module.new("E")
+      scope3 << scope4
+      scope2 << scope3
+      scope1 << scope2
+      rbi << scope1
+      rbi << scope5
```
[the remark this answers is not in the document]
jeffcarbs: That's way more readable 🙂 - updated.

test/rbi/rewriters/flatten_scopes_test.rb
jeffcarbs: Added 👍

lib/rbi/rewriters/flatten_scopes.rb
jeffcarbs: I tried this type of approach and couldn't get it to work. Where you have `parent_tree` you actually need "grandparent" (i.e. `node.parent_tree&.parent_tree`) since we're flattening by removing a node from its parent, adding the namespace to the name, and reattaching to the grandparent. Even then, I couldn't exactly figure out why (got a bit lost in the recursion) but I think it has to do with the order in which you're visiting and detaching/reattaching the nodes.

You're welcome to play around with it yourself and push a commit if you can get it working, but it's pretty terse as-is so I think it may be OK.

test/rbi/rewriters/flatten_scopes_test.rb
```diff
@@ -0,0 +1,186 @@
...
+      scope3 = RBI::Class.new("C")
+      scope3 << RBI::Const.new("C1", "42")
+      scope4 = RBI::Module.new("D")
+      scope5 = RBI::Module.new("E")
+      scope5 << RBI::Const.new("E1", "42")
+      scope3 << scope4
+      scope2 << scope3
+      scope1 << scope2
+      rbi << scope1
+      rbi << scope5
```
Morriar, line 64: I don't think we need this anymore

```suggestion
```

jeffcarbs: Whoops, missed that. Just fixed-up that commit and force-pushed.

PR #114, opened 2021-12-20
Handle AttrAccessor/Reader/Writer in GroupNodes

jeffcarbs:
Came across a runtime error trying to use `AttrReader` in a custom DSL generator:
> RuntimeError: Unknown group for .attr_reader(:foo)

This PR handles attrs in GroupNodes. For now I've just added them to the `Sends` group but open question:
- Should these be in their own "kind" - e.g. `Group::Kind::Attrs`
- If so, where should they go inside `SortNodes`? I would think maybe right below Sends?

### DIFF
--- a/lib/rbi/rewriters/sort_nodes.rb
+++ b/lib/rbi/rewriters/sort_nodes.rb
@@ -8,6 +8,7 @@ module RBI
 
       sig { override.params(node: T.nilable(Node)).void }
       def visit(node)
+        sort_node!(node) if node
         return unless node.is_a?(Tree)
         visit_all(node.nodes)
         original_order = node.nodes.map.with_index.to_h
@@ -38,9 +39,7 @@ module RBI
         when Send                 then 50
         when TStructField         then 60
         when TEnumBlock           then 70
-        when AttrAccessor         then 71
-        when AttrReader           then 72
-        when AttrWriter           then 73
+        when Attr                 then 75
         when Method
           if node.name == "initialize"
             81
@@ -81,6 +80,16 @@ module RBI
         case node
         when Module, Class, Struct, Const, Method, Helper, TStructField
           node.name
+        when Attr
+          node.names.first.to_s
+        end
+      end
+
+      sig { params(node: Node).void }
+      def sort_node!(node)
+        case node
+        when Attr
+          node.names.sort!
         end
       end
     end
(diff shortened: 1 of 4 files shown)


Morriar (commented):
Good catch 👍 

Let's create a new group that would go just before the `Method`.

jeffcarbs:
@Morriar - Made `Attrs` its own group and put it above `Methods`. I also ordered the attrs themselves so `accessor, reader, writer`. I just did alphabetically, although for some reason `reader, writer, accessor` seems more natural to me I think? It probably doesn't matter either way, but I'll defer to you.

jeffcarbs:
@Morriar - wanted to give this a post-holiday bump 🙂 lmk if there are more adjustments you’d like to see.

jeffcarbs:
What about `attr_*` with multiple names - I assume we'd just take the first name? They may not necessarily be sorted, so you could wind up with this:
```ruby
attr_accessor :foo_accessor, :bar_accessor, :baz_accessor
attr_reader :bar_reader, :baz_reader, :foo_reader
attr_writer :baz_writer, :foo_writer, :bar_writer
```

becoming
```ruby
attr_reader :bar_reader, :baz_reader, :foo_reader
attr_writer :baz_writer, :foo_writer, :bar_writer
attr_accessor :foo_accessor, :bar_accessor, :baz_accessor
```

which seems sort of arbitrary but then again so does probably any decision we make here 😅 . Lmk what you think, happy to change the code to get this landed. It's blocking the conversion of the last manual RBI generator we have into a custom tapioca dsl compiler.

jeffcarbs:
Makes sense, I can take a stab at sorting the attr names.

lib/rbi/rewriters/sort_nodes.rb
```diff
@@ -8,6 +8,7 @@ class SortNodes < Visitor
 
       sig { override.params(node: T.nilable(Node)).void }
       def visit(node)
+        sort_node!(node) if node
```
Morriar, line 11:
```suggestion
        sort_node!(node) if node
        
```

Morriar (commented):
Almost good to go 🎉

lib/rbi/rewriters/sort_nodes.rb
jeffcarbs:
```suggestion
        sort_node_names!(node) if node

```

jeffcarbs:
Thanks for the review, I applied the suggested changes 👍

Morriar (approved):
Thank you for pushing this through!

PR #116, opened 2022-01-27
Pass max_line_length to printer

Morriar:
This option is available on `Node#print` it should also be on `File#print`.

### DIFF
--- a/lib/rbi/printer.rb
+++ b/lib/rbi/printer.rb
@@ -117,16 +117,23 @@ module RBI
       end
     end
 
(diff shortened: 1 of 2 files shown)


Morriar:
> What does it mean for a file to have a max line length? Does it mean all lines in the file are wrapped at the limit?

Yes, the file call recursively `print` on its own node with the `max_line_length` resulting on the sig blocks to be wrapped if too long.

vinistock (approved):

KaanOzkan (approved):

PR #118, opened 2022-01-27
Do not try to instanciate abstract class Visibility

Morriar:
We'll have to check why Sorbet doesn't see this.

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -289,7 +289,14 @@ module RBI
         names = node.children[2..-1].map { |child| parse_name(child) }
         MixesInClassMethods.new(*names, loc: loc, comments: comments)
       when :public, :protected, :private
-        visibility = Visibility.new(method_name, loc: loc)
+        visibility = case method_name
+        when :protected
+          Protected.new(loc: loc)
+        when :private
+          Private.new(loc: loc)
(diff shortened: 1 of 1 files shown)


KaanOzkan (approved):
Probably because this definition in `printer` is not abstract? https://github.com/Shopify/rbi/blob/903a2f73e6813619481fa08b6c7b11045b8123f7/lib/rbi/printer.rb#L573

vinistock:
Do we still need `Visibility#initialize`? Should it even exist? https://github.com/Shopify/rbi/blob/5033e1bac80d19126ff278a31680ad785cd1470f/lib/rbi/model.rb#L843

Morriar:
> Probably because this definition in `printer` is not abstract?

I don't think so. It doesn't seem to be checked either on a [simpler case](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3AHelpers%0A%0A%20%20abstract!%0Aend%0A%0AFoo.new).

Morriar:
> Do we still need `Visibility#initialize`? Should it even exist?

Yes, we still need Visibility to set its internal state (@visivility) and call super. Sorbet should prevent calling `.new` tho since it's abstract.

PR #120, opened 2022-01-28
Add support for final sigs

paracycle:
Add support for parsing and writing `sig(:final)` calls which represent final methods: https://sorbet.org/docs/final#final-methods

### DIFF
--- a/lib/rbi/printer.rb
+++ b/lib/rbi/printer.rb
@@ -670,7 +670,9 @@ module RBI
 
     sig { params(v: Printer).void }
     def print_as_line(v)
-      v.printt("sig { ")
+      v.printt("sig")
+      v.print("(:final)") if is_final
+      v.print(" { ")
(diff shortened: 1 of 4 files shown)


lib/rbi/printer.rb
```diff
@@ -670,7 +670,15 @@ def sig_modifiers
...
     def print_as_line(v)
-      v.printt("sig { ")
+      sig_modifier = if is_final
+        "(:final)"
+      else
+        ""
+      end
+
+      v.printt("sig")
+      v.print(sig_modifier)
```
Morriar, line 680: Can we do this instead?

```rb
v.print("(:final)") if is_final
```

paracycle: Haha. Yeah, makes sense. I think I was thinking that there could different modifier parameters, but it turns out `:final` is the only one anyway.

Morriar (approved):

PR #122, opened 2022-02-26
Missing T::Struct options that impact type checking

jeffcarbs:
We started using the `factory` option for a field in one of our T::Structs and noticed it didn't come through the RBI that tapioca generated. `factory` is basically the same as `default` but takes a proc so the value can be set dynamically. Like default, if a prop/const has a factory set you don't have to pass a value in the initializer so its absence from the RBI meaningfully impacts type-checking. Given that tapioca uses the `RBI::TStructConst/Prop` models, support needs to be added here first.

I started down the path of a fix, adding handling for `factory` where there's handling for `default`, when I realized that there's at least one more option that affects type checking: `immutable`. If you set `prop :x, Integer, immutable: true` it basically acts like `const` where it will show a type error if you try to use the setter. I also realized that the current parsing logic expects `default` (if present at all) to be the first option so if you have `immutable: true, default: 1` the parser will currently drop both on the floor.

Here's a [sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20Foo%20%3C%20T%3A%3AStruct%0A%20%20const%20%3Ax%2C%20Integer%2C%20factory%3A%20-%3E%20%7B%7D%0A%20%20const%20%3Ay%2C%20Integer%2C%20factory%3A%20T.unsafe%28nil%29%0A%20%20prop%20%3Az%2C%20Integer%2C%20immutable%3A%20true%2C%20some_other_prop%3A%20'hey'%0Aend%0A%0AFoo.new%28z%3A%201%29.z%20%3D%202) to showing these examples.

There are [many more options](https://github.com/sorbet/sorbet/blob/c3365c5fbc4f652b30cc45428792f4646783c3a8/gems/sorbet-runtime/lib/types/props/_props.rb#L54-L109) (although many seem to be internal to stripe) and technically you can pass arbitrary options.

Before I went further I wanted to get your thoughts. I think we at least need to handle `factory` and `immutable` since those impact type checking, so there are a few options:
- add `factory` and `immutable` attrs to `TStructField` (and const/prop) and handle them whenever we currently handle `default`.
  - Slightly more verbose but very explicit
  - We'll drop other options on the floor. This is probably fine but it would mean that `input == (parse(input).out)` is no longer necessarily true.
- remove `default` from `TStructField` and replace it with `options` which would capture all options
  - More flexible but less explicit
- just add factory, and when we see “immutable: true” parse it as a Const instead of a Prop.

There's also a middle ground where the constructor for `TStructField` and friends still only take `default` (and maybe `factory` and `immutable` as well?) as an explicit option since it's the most common but also take an `options` hash. The initializer would join the two so you'd only ever operate on the hash. This would be similar to how `Attr` takes (name, names) and then joins them together in the initializer.

I'm happy to take this on here and in tapioca so let me know what you think.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -1142,23 +1142,23 @@ module RBI
     sig { returns(String) }
     attr_accessor :name, :type
 
-    sig { returns(T.nilable(String)) }
-    attr_accessor :default
+    sig { returns(T::Hash[Symbol, String]) }
+    attr_accessor :options
 
     sig do
       params(
         name: String,
         type: String,
-        default: T.nilable(String),
+        options: T::Hash[Symbol, String],
         loc: T.nilable(Loc),
         comments: T::Array[Comment]
       ).void
     end
-    def initialize(name, type, default: nil, loc: nil, comments: [])
+    def initialize(name, type, options: {}, loc: nil, comments: [])
       super(loc: loc, comments: comments)
       @name = name
       @type = type
-      @default = default
+      @options = options
     end
 
     sig { abstract.returns(T::Array[String]) }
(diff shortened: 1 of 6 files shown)


Morriar (commented):
Thanks for taking a look at this 🙏 

I like the approach, my only concern is the change in the constructor signature for `TStructField` which will be breaking compatibility.

PR #123, opened 2022-03-03
Fix parsing enum with one value

jeffcarbs:
The AST for the `enums do ... end` block is different when there's a single enum variant vs when there are multiple. The parser was blowing up for the single-variant case, this PR fixes that.

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -429,7 +429,14 @@ module RBI
     sig { params(node: AST::Node).returns(TEnumBlock) }
     def parse_enum(node)
       enum = TEnumBlock.new
-      node.children[2].children.each do |child|
+
+      body = if node.children[2].type == :begin
+        node.children[2].children
+      else
+        [node.children[2]]
+      end
+
+      body.each do |child|
         enum << parse_name(child)
       end
       enum.loc = node_loc(node)


lib/rbi/parser.rb
```diff
@@ -429,7 +429,14 @@ def parse_sig(node)
     sig { params(node: AST::Node).returns(TEnumBlock) }
     def parse_enum(node)
       enum = TEnumBlock.new
-      node.children[2].children.each do |child|
+
+      if node.children[2].type == :begin
```
jeffcarbs, line 433: There's a similar thing being done in the `T::Struct` parsing code: https://github.com/Shopify/rbi/blob/03d02e24f0dc768cca68aeb3f600dc209727a514/lib/rbi/parser.rb#L377-L381

lib/rbi/parser.rb
```diff
@@ -429,7 +429,14 @@ def parse_sig(node)
...
+        body = node.children[2].children
+      else
+        body = [node.children[2]]
+      end
```
Morriar, line 437:
```suggestion
      body =  if node.children[2].type == :begin
        node.children[2].children
      else
        [node.children[2]]
      end
```

Morriar (commented):
Nice find, I didn't think about enums with only one value.

For my own curiosity, what's the point such enums?

jeffcarbs:
> For my own curiosity, what's the point such enums?

We have two such cases:
-  one is a case where we have one for now and intend to add more over time.
- Another is a enum that has a bunch of other sibling enum classes. The siblings all have many variants but this one currently only has one. We have a big T.any type alias that enumerates all of these enum classes so to make things work seamlessly this has to quack like the rest even though it could very well be a single constant.
  - similar to the other case, this also allows adding a new enum variant without a big refactor.

Morriar:
@jeffcarbs I released [v0.0.13](https://rubygems.org/gems/rbi/versions/0.0.13) with this patch 👍

PR #142, opened 2022-12-23
Do not change visibility when sorting

### LINKED ISSUE #141: `sort_nodes!` changes method privacy
I think that the `SortNodes` rewriter doesn't preserve the position of `private` and `protected`, causing it to change the privacy of methods. For example, I would expect this input:

```ruby
rbi = RBI::Tree.new
rbi << RBI::Method.new("c")
rbi << RBI::Private.new
rbi << RBI::Method.new("a")
rbi << RBI::Protected.new
rbi << RBI::Method.new("b")
rbi.sort_nodes!
puts rbi.string
```

to produce
```ruby
def c; end
private
def a; end
protected
def b; end
```

but it sorts everything and produces
```ruby
def a; end
def b; end
def c; end
private
protected
```

So that method `b` which should be private has become public, and `c` which should be protected is also public.

If my understanding is correct, the problem is in the `node.nodes.sort!` call in the `SortNodes` visitor, which currently sorts all the nodes. I think it needs to first segment the nodes into public, private, and protected ranges and sort those ranges individually. Would a PR to fix this be helpful?

mwudka:
Hopefully fixes #141 . Updates the `SortNodes` visitor to preserve visibility of nodes. So, a file that looks like

```ruby
class MyClass
def zethod2; end
def zethod1; end
private
def c; end
def b; end
end
```

will get sorted as
```ruby
class MyClass
def zethod1; end
def zethod2; end
private
def b; end
def c; end
end
```

rather than as

```ruby
class MyClass
def b; end
def c; end
private
def zethod1; end
def zethod2; end
end
```

### DIFF
--- a/lib/rbi/rewriters/sort_nodes.rb
+++ b/lib/rbi/rewriters/sort_nodes.rb
@@ -13,18 +13,29 @@ module RBI
         return unless node.is_a?(Tree)
         visit_all(node.nodes)
         original_order = node.nodes.map.with_index.to_h
-        node.nodes.sort! do |a, b|
-          # First we try to compare the nodes by their node rank (based on the node type)
-          res = node_rank(a) <=> node_rank(b)
-          next res if res != 0 # we can sort the nodes by their rank, let's stop here
 
-          # Then, if the nodes ranks are the same (res == 0), we try to compare the nodes by their name
-          res = node_name(a) <=> node_name(b)
-          next res if res && res != 0 # we can sort the nodes by their name, let's stop here
+        # The child nodes could contain private/protected markers. If so, they should not be moved in the file.
+        # Otherwise, some methods could see their privacy change. To avoid that problem, divide the array of child
+        # nodes into chunks based on whether any Visibility nodes appear, and sort the chunks independently. This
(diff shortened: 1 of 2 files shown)


KaanOzkan (commented):
Thanks!

I like the 2 pass solution. Although it was hard to understand at first, it may be useful to extract the core sorting logic to its own method (L37-L46). What do you think?

paracycle:
I think there is a better way to do this by using `partition` and sorting each of the partitions independently. If we did that, we wouldn't have to do all the manual offset calculations that are present in this PR.

mwudka:
After looking into `partition`, I realized that my original approach was wrong: visibility modifiers can be repeated, so there can be an arbitrary number of chunks. The helpful `partition` suggestion from @paracycle got me looking into other ruby array methods I don't know, and I found `chunk` which I think makes things much clearer.

It does increase allocations, but perhaps that is a good tradeoff here, since this isn't runtime code?

@KaanOzkan the code is now much shorter, so perhaps the helper method is not needed? I am on the fence, but leaned towards keeping it inline so that the flow is clear and matches the existing structure more closely.

KaanOzkan (approved):
This is definitely simpler thank you!

PR #145, opened 2023-03-29
Update Rubocop

egiurleo:
I was running into the following error from VS Code when developing on the RBI repo:

```
RuboCop found unknown Ruby version 3.1 in `.ruby-version`. Supported versions: 2.5, 2.6, 2.7, 3.0
```

Updating the Rubocop gem fixes this issue.


andyw8:
Should Dependabot be enabled for this repo? I see only [one previous PR](https://github.com/Shopify/rbi/pull/57).

vinistock:
We should also add dependabot to this repo.

PR #160, opened 2023-04-04
Bump Sorbet version and introduce AbstractParam

st0012:
Currently, the `Param` class plays 2 roles:

1. Being a class that represents normal params (used in projects like Tapioca)
2. Being the abstract class for other param types

However, after https://github.com/sorbet/sorbet/pull/6273 Sorbet stops prohibiting instantiating abstract classes, which means these 2 roles can't co-exist on the same calss anymore.

I think that leaves us with 2 options:

1. Create an `AbstractParam` class (I'm open for a better name) as all param classes' (including `Param`'s) superclass
2. Or create a new class to present normal params and migrate `Param` usages to it

Since 2) will require downstream projects to adopt new a new param class, it's functionally a breaking change.
And although 1) will also require changes, but it'll be on signatures only, which IMO can be argued as NOT a breaking change?

WDYT?

### Required changes on Tapioca

```diff
diff --git a/lib/tapioca/helpers/rbi_helper.rb b/lib/tapioca/helpers/rbi_helper.rb
index 4796de77..f0fc1d7d 100644
--- a/lib/tapioca/helpers/rbi_helper.rb
+++ b/lib/tapioca/helpers/rbi_helper.rb
@@ -73,7 +73,7 @@ module Tapioca
       create_typed_param(RBI::BlockParam.new(name), type)
     end
 
-    sig { params(param: RBI::Param, type: String).returns(RBI::TypedParam) }
+    sig { params(param: RBI::AbstractParam, type: String).returns(RBI::TypedParam) }
     def create_typed_param(param, type)
       RBI::TypedParam.new(param: param, type: sanitize_signature_types(type))
     end
diff --git a/lib/tapioca/rbi_ext/model.rb b/lib/tapioca/rbi_ext/model.rb
index a85bb1d6..6aedc3c2 100644
--- a/lib/tapioca/rbi_ext/model.rb
+++ b/lib/tapioca/rbi_ext/model.rb
@@ -124,7 +124,7 @@ module RBI
   end
 
   class TypedParam < T::Struct
-    const :param, RBI::Param
+    const :param, RBI::AbstractParam
     const :type, String
   end
 end
```


Morriar (changes requested):
I don't think the change should be done here. `Param` is already abstract so the fault is on whatever is trying to instantiate it.

From what I understand the problem is mostly because of this line in Tapioca: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/helpers/rbi_helper.rb#L43 ? Should it be using `ReqParam` instead?

st0012:
@Morriar Yeah I think that makes more sense. I've opened https://github.com/Shopify/tapioca/pull/1467 for it.

st0012:
Close in favour of https://github.com/Shopify/rbi/pull/161

PR #180, opened 2023-05-19
Proof of concept: Support inline RBI versioning

egiurleo:
This is a proof of concept for RBI versioning, which @Morriar, @bitwise-aiden, and I had been working on for fun a while back.

## Problem
This proof of concept aims to solve the problem of versioning RBIs. Up until now, there is no default way to write RBI for different versions of the same gem. This can lead to confusing issues for Sorbet users where RBIs don't match the version of a gem they're using, and they have to spend time figuring out whether their gem is out of date, whether the RBI is wrong, or if it's even both! Having a way to specify RBI for multiple versions of the same gem would allow people to adopt and use Sorbet without having to go through this hassle.

## Approach

To implement this behavior, we decided to create a system of versioning RBI within the same file using `@version` annotations within comments. This is an example of using such an annotation:

```ruby
class Foo

  # @version > 0.3.0
  def bar
  end

  # @version <= 0.3.0
  def bar(arg1)
  end
end
```

In this example, the class `Foo` has a method `bar` that takes one argument in versions 0.3.0 and earlier, but that argument is removed in later versions.

There are a few benefits to this approach:
- Limits duplication by preventing multiple files being created for the same gem RBI
- Less duplication means that relevant annotations are easier to find and read all at once
- When parsing the RBI files, parts that are not relevant to the current gem version can be removed

There are also some drawbacks, namely that RBI files could potentially become cluttered if many gem versions are supported. We plan to address this using some automated linting (see "Future Work" section), and in practice we don't believe that most gem authors actually support that many versions at a time. Another potential drawback is that parsing out irrelevant parts of the RBI can slow down parsing, but in practice, we see that most RBI files are relatively small and that speed is not currently a big issue; however, this can be revisited if performance becomes a bigger priority.

For more information about the other approaches we considered, see [this document](https://docs.google.com/document/d/1IlffQ_IiFMi2XrNRI3jxhHOiZaBKaVZ6xJx0RHBIIw0/edit#heading=h.edz4gygnhpur).

## Implementation

This change was implemented by creating a new rewriter called `FilterVersions`. This takes the current RBI tree as well as a gem version to filter on. For each node in the RBI, the rewriter determines if the node satisfies the gem version by examining any `@version` annotations in the comments and relying on logic from Ruby's `Gem::Requirement` class to perform comparisons.

RBIs with no versions remain unaffected.

## Future Work

1. Documentation
If the team is supportive of this approach, I'll write up documentation for this behavior before merging the PR.

2. Linting
While we believe the inline versioning approach makes the most sense for RBI, it still leaves the possibility of developers writing disorganized RBIs (e.g. different versions of the same method are far away from each other) or invalid RBI files (e.g. writing versions that are somehow incompatible with each other). To address these issues, we would implement some sort of linting that prevents developers from falling into these mistakes so that this feature is as easy to use as possible.


lib/rbi/rewriters/filter_versions.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  class Node
+    sig { params(version: Gem::Version).returns(T::Boolean) }
+    def satisfies_version?(version)
+      return true unless is_a?(NodeWithComments)
+
+      requirements = version_requirements
+      requirements.empty? || requirements.any? { |req| req.satisfied_by?(version) }
```
egiurleo, line 47: Note: any node with no requirements is treated as satisfying the requirements.

test/rbi/rewriters/filter_versions_test.rb
```diff
@@ -0,0 +1,284 @@
...
+        class Bar; end
+
+        # @version >= 0.4.0-prerelease
+        class Buzz; end
+      RBI
+    end
+
+    def test_filter_versions_and
+      rbi = <<~RBI
+        # @version > 0.3.0, < 1.0.0
```
egiurleo, line 194: Note: this implementation supports and/or logic. The line above is an example of "and" -- the version must be greater than 0.3.0 AND less than 1.0.0. Line 218 is an example of "or" logic.

bitwise-aiden: > I wonder if only permitting the same syntax as a `.gemspec` would be feasible
Could you please elaborate on this a bit more @andyw8?

lib/rbi/rewriters/filter_versions.rb
```diff
@@ -0,0 +1,112 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RBI
+  module Rewriters
+    # Take a gem version and filter out all RBI that is not relevant to that version based on @version annotations
```
egiurleo, line 6: @Morriar Can you take a look at this comment and let me know what you think!

Morriar: The comment is really good 👍 

No need for an experimental mention, `rbi` just provides the rewriter, any tool that builds a feature on top of this may in the other hand mark it as experimental.

lib/rbi/rewriters/filter_versions.rb
```diff
@@ -0,0 +1,113 @@
...
+    #     def bar
+    #     end
+    #
+    #     # @version <= 0.3.0
+    #     def bar(arg1)
+    #     end
+    #   end
+    # RBI
+    #
+    # Rewriters::FilterVersions.filter(tree, Gem::Version.new("0.3.1"))
```
Morriar, line 22: I think we should provide a `Tree#filter_versions!` method like we do on other rewriter (see the sorting for example: https://github.com/Shopify/rbi/blob/main/lib/rbi/rewriters/sort_nodes.rb#L115-L123).

Morriar (approved):

PR #195, opened 2023-06-05
Method nodes fail tree merge if their default values are not exactly the same

paracycle:
@searls reported that Tapioca would (silently) fail to merge the generated signatures and the bundled signatures of a method defined like the following:
```ruby
sig { params(a: T::Boolean).void }
def foo(a: false)
end
```

This happens because when Tapioca generates the RBI for the `foo` method it generates it as:
```ruby
sig { params(a: T::Boolean).void } 
def foo(a: T.unsafe(nil); end
```

Note that the default value in the generated RBI file does not match the default value in the bundled RBI file, so the merge fails [this check](https://github.com/Shopify/rbi/blob/da5e30e95f290aa26fbdcf5fc9fcace5c5a291aa/lib/rbi/rewriters/merge_trees.rb#L464), which itself fails because of [this comparison](https://github.com/Shopify/rbi/blob/36de54c34409292f9fab8e7c4a3a482adfc756ab/lib/rbi/model.rb#L708-L711).

This PR adds a test case to exhibit the behaviour and adds a hacky fix that considers `T.unsafe(nil)` to be equivalent as a default value to any other value. We probably need a more general fix for this though.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -621,7 +621,7 @@ module RBI
 
     sig { params(other: T.nilable(Object)).returns(T::Boolean) }
     def ==(other)
-      OptParam === other && name == other.name && value == other.value
+      OptParam === other && name == other.name
     end
   end
 
@@ -707,7 +707,7 @@ module RBI
 
     sig { params(other: T.nilable(Object)).returns(T::Boolean) }
     def ==(other)
-      KwOptParam === other && name == other.name && value == other.value
+      KwOptParam === other && name == other.name
     end
   end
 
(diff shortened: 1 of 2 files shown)


Morriar (approved):
Do we actually care about the default value that much?

I wonder if we could merge even if it's something different than `T.unsafe`?

paracycle:
Ok, removed the comparison of `value`s when comparing `OptParam` and `KwOptParam` nodes. I had to change 1 test that was assuming that a different default value would create a merge conflict, which in the new reality it doesn't.

searls:
Thank you!

PR #233, opened 2023-09-12
Generate RBI signatures for YARP

Morriar:
This PR adds a script under `bin/yarp-rbi` that generates the signatures for all YARP nodes using the current version of the `yarp` gem.

This will make it easier to maintain them up-to-date after each bump.


bin/yarp-rbi
```diff
@@ -0,0 +1,162 @@
...
+YARP_GEM = "yarp"
+YARP_CONFIG = "config.yml"
+
+RBI_HEADER = T.let(<<~RBI, String)
+  # typed: true
+
+  # DO NOT EDIT MANUALLY
+  # This is an autogenerated file for YARP, run `bin/yarp-rbi` instead.
+
+  module YARP
```
vinistock, line 20: You're missing the visitor methods here, which could be useful.

PR #240, opened 2023-10-16
Migrate from YARP to Prism

vinistock:
This PR migrates YARP to Prism. We need to migrate all our gems to use Prism because having a mix of YARP and Prism in the bundle can lead to unexpected behaviour - since YARP v0.13 also exports a default require named `prism`.

This PR

- Changes the runtime dependency to be Prism instead of YARP
- Bumps the Ruby version (we were still developing using Ruby 2.7)
- Fixes the breaking changes related to the switch


st0012 (approved):

lib/rbi/parser.rb
```diff
@@ -156,32 +156,25 @@ class TreeBuilder < Visitor
...
-        return unless node
-
-        @last_node = node
-        super
-      end
-
-      sig { override.params(node: YARP::ClassNode).void }
+      sig { override.params(node: Prism::ClassNode).void }
       def visit_class_node(node)
+        @last_node = node
```
Morriar, line 177: Since we're now only setting `last_node` for the node we have an explicit visit on, should we set it to `nil` before leaving the node to avoid errors pointing to the wrong node?

vinistock: Good point! Done

Morriar (approved):

PR #241, opened 2023-10-26
Bump Prism dependency

andyw8:
Fixes a compilation error that was preventing us from releasing an update to the gem.


andyw8:
(should we release as v0.1.3?)

vinistock:
Let's re-generate the RBIs in case there are typechecking errors. I think we should be good to ship the new version.

vinistock (approved):

PR #242, opened 2023-10-26
Move Prism shim to correct location and format with RuboCop

andyw8:


sorbet/rbi/shims/gems/prism.rbi
```diff
@@ -1,11 +1,9 @@
 # typed: strict
 # frozen_string_literal: true
 
-=begin
-This file is generated by the templates/template.rb script and should not be
-modified manually. See templates/rbi/prism.rbi.erb
-if you are looking to modify the template
-=end
+# This file is generated by the templates/template.rb script and should not be
```
andyw8, line 4: Not actually true...

vinistock (approved):

PR #296, opened 2024-05-02
Remove Ruby 3.0 from CI matrix

Morriar:
Ruby 3.0 is EOL since 2024-04-23.

See https://www.ruby-lang.org/en/downloads/branches/.


andyw8 (approved):

Morriar:
I also bumped the require Ruby version in the Gemspec

PR #298, opened 2024-05-03
Remove upper limit constraint on prism

KaanOzkan:


eregon:
Some overlap with https://github.com/Shopify/rbi/pull/297

PR #299, opened 2024-05-16
Set dependabot `open-pull-requests-limit` to 100

egiurleo:


.github/dependabot.yml
```diff
@@ -3,7 +3,7 @@ updates:
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
-      interval: "monthly"
+      interval: "weekly"
```
andyw8, line 6: 🤔 Could we opt-out and keep it as monthly? (cc @KaanOzkan since you changed it in https://github.com/Shopify/rbi/pull/220).

dirceu (approved):

PR #308, opened 2024-06-06
Add rewriter to transform attributes into methods

amomchilov:
Part of https://github.com/Shopify/tapioca/issues/1918

This PR introduces a new rewriter which transforms `attr_reader`/`attr_writer`/`attr_accessor` RBIs into methods with the corresponding sigs. This has two benefits:

1. It naturally solves Shopify/tapioca#1709. If Tapioca only has to deal with methods (which can already be merged), then we don't have to write bespoke logic to merge attributes with methods, or methods with attributes.
2. It'll leave less work for Sorbet, which internally rewrites attributes into methods. We can do this on its behalf, once up-front, rather than on every parse of the RBIs.

### DIFF
--- a/lib/rbi/rewriters/attr_to_methods.rb
+++ b/lib/rbi/rewriters/attr_to_methods.rb
@@ -1,0 +1,169 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RBI
+  class UnexpectedMultipleSigsError < Error
+    sig { returns(Node) }
+    attr_reader :node
+
+    sig { params(node: Node).void }
+    def initialize(node)
+      super(<<~MSG)
+        This declaration cannot have more than one sig.
+
+        #{node.string.chomp}
+      MSG
+
+      @node = node
+    end
+  end
+
+  module Rewriters
+    class AttrToMethods < Visitor
+      extend T::Sig
+
+      sig { override.params(node: T.nilable(Node)).void }
+      def visit(node)
+        case node
+        when Tree
+          visit_all(node.nodes.dup)
+
+        when Attr
+          replace(node, with: node.convert_to_methods)
+        end
+      end
+
+      private
+
+      sig { params(node: Node, with: T::Array[Node]).void }
+      def replace(node, with:)
+        tree = node.parent_tree
+        raise ReplaceNodeError, "Can't replace #{self} without a parent tree" unless tree
+
+        node.detach
+        with.each { |node| tree << node }
+      end
+    end
+  end
+
+  class Tree
+    extend T::Sig
+
+    sig { void }
+    def replace_attributes_with_methods!
+      visitor = Rewriters::AttrToMethods.new
+      visitor.visit(self)
+    end
+  end
+
+  class Attr
+    sig { abstract.returns(T::Array[Method]) }
+    def convert_to_methods; end
+
+    private
+
+    sig(:final) { returns([T.nilable(Sig), T.nilable(String)]) }
+    def parse_sig
+      raise UnexpectedMultipleSigsError, self if 1 < sigs.count
+
+      sig = sigs.first
+      return [nil, nil] unless sig
+
+      attribute_type = case self
+      when AttrReader, AttrAccessor then sig.return_type
+      when AttrWriter then sig.params.first&.type
+      end
+
+      [sig, attribute_type]
+    end
+
+    sig do
+      params(
+        name: String,
+        sig: T.nilable(Sig),
+        visibility: Visibility,
+        loc: T.nilable(Loc),
+        comments: T::Array[Comment],
+      ).returns(Method)
+    end
+    def create_getter_method(name, sig, visibility, loc, comments)
+      Method.new(
+        name,
+        params: [],
+        visibility: visibility,
+        sigs: sig ? [sig] : [],
+        loc: loc,
+        comments: comments,
+      )
+    end
+
+    sig do
+      params(
+        name: String,
+        sig: T.nilable(Sig),
+        attribute_type: T.nilable(String),
+        visibility: Visibility,
+        loc: T.nilable(Loc),
+        comments: T::Array[Comment],
+      ).returns(Method)
+    end
+    def create_setter_method(name, sig, attribute_type, visibility, loc, comments) # rubocop:disable Metrics/ParameterLists
+      sig = if sig # Modify the original sig to correct the name, and remove the return type
+        params = attribute_type ? [SigParam.new(name, attribute_type)] : []
+
+        Sig.new(
+          params: params,
+          return_type: "void",
+          is_abstract: sig.is_abstract,
+          is_override: sig.is_override,
+          is_overridable: sig.is_overridable,
+          is_final: sig.is_final,
+          type_params: sig.type_params,
+          checked: sig.checked,
+          loc: sig.loc,
+        )
+      end
+
+      Method.new(
+        "#{name}=",
+        params: [ReqParam.new(name)],
+        visibility: visibility,
+        sigs: sig ? [sig] : [],
+        loc: loc,
+        comments: comments,
+      )
+    end
+  end
+
+  class AttrAccessor
+    sig { override.returns(T::Array[Method]) }
+    def convert_to_methods
+      sig, attribute_type = parse_sig
+
+      names.flat_map do |name|
+        [
+          create_getter_method(name.to_s, sig, visibility, loc, comments),
+          create_setter_method(name.to_s, sig, attribute_type, visibility, loc, comments),
+        ]
+      end
+    end
+  end
+
+  class AttrReader
+    sig { override.returns(T::Array[Method]) }
+    def convert_to_methods
+      sig, _ = parse_sig
+
+      names.map { |name| create_getter_method(name.to_s, sig, visibility, loc, comments) }
+    end
+  end
+
+  class AttrWriter
+    sig { override.returns(T::Array[Method]) }
+    def convert_to_methods
+      sig, attribute_type = parse_sig
+
+      names.map { |name| create_setter_method(name.to_s, sig, attribute_type, visibility, loc, comments) }
+    end
+  end
+end
--- a/lib/rbi.rb
+++ b/lib/rbi.rb
@@ -5,7 +5,9 @@ require "sorbet-runtime"
 require "stringio"
 
 module RBI
-  class Error < StandardError; end
+  class Error < StandardError
+    extend T::Sig
+  end
 end
 
 require "rbi/loc"
@@ -21,6 +23,7 @@ require "rbi/rewriters/nest_singleton_methods"
 require "rbi/rewriters/nest_non_public_methods"
 require "rbi/rewriters/group_nodes"
 require "rbi/rewriters/remove_known_definitions"
+require "rbi/rewriters/attr_to_methods"
 require "rbi/rewriters/sort_nodes"
 require "rbi/parser"
 require "rbi/printer"
--- a/test/rbi/rewriters/attr_to_methods_test.rb
+++ b/test/rbi/rewriters/attr_to_methods_test.rb
@@ -1,0 +1,311 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module RBI
+  class AttrToMethodsTest < Minitest::Test
+    def test_replaces_attr_reader_with_method
+      rbi = Parser.parse_string(<<~RBI)
+        # Lorum ipsum...
+        sig { returns(Integer) }
+        attr_reader :a
+      RBI
+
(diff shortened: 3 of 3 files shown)


lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,147 @@
...
+      )
+
+      [m]
+    end
+  end
+
+  class AttrWriter
+    sig { returns(T::Array[Method]) }
+    def convert_to_methods
+      raise "How can there be multiple?" unless sigs.one?
```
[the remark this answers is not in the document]
amomchilov: Not for attributes, though. [Sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20returns%28Integer%29%20%7D%0Asig%20%7B%20returns%28Integer%29%20%7D%0Aattr_reader%20%3Afoo%0A)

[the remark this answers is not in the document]
Morriar: In `.rbi` files only. With sorbet.run you're testing with a `.rb` file.

amomchilov:
Spoke offline. Conclusions:

* Q1: Be flexible, and accept both `.void` and `.returns(Type)`. Reevaluate if Sorbet changes its behaviour to prefer one or the other
* Q2: Always emit `.void`.

lib/rbi/rewriters/replace_attributes_with_methods.rb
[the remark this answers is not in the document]
amomchilov: Indeed, looks like it allows it, but the revealed type is whichever one you listed first. I think we should warn in that case, it likely wasn't intended.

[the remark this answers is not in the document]
Morriar: We currently do not do any linting / warning inside `rbi`, it's up to the client to make sure that what is produced is correct.

This could be a warning in Tapioca when pulling the gem's hand-written annotations. Or it could be done by Sorbet.

lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,146 @@
...
+        params(
+          name: String,
+          sig: T.nilable(Sig),
+          attribute_type: T.nilable(String),
+          visibility: Visibility,
+          loc: T.nilable(Loc),
+          comments: T::Array[Comment],
+        ).returns(Method)
+      end
+      def create_setter_method(name, sig, attribute_type, visibility, loc, comments) # rubocop:disable Metrics/ParameterLists
```
amomchilov, line 119: 6 params, but the upper limit is 5. Worth overriding?

test/rbi/rewriters/replace_attributes_with_methods_test.rb
```diff
@@ -0,0 +1,292 @@
...
+        def a=(a); end
+
+        # Lorum ipsum...
+        sig { params(b: Integer).void }
+        def b=(b); end
+
+        # Lorum ipsum...
+        sig { params(c: Integer).void }
+        def c=(c); end
+      RBI
```
amomchilov, line 162: On second thought, I think it might be better to group by property name, rather than by reader vs writer, i.e.:

```ruby
        sig { returns(Integer) }
        def a; end

        sig { params(a: Integer).void }
        def a=(a); end

        sig { returns(Integer) }
        def b; end

        sig { params(b: Integer).void }
        def b=(b); end

        sig { returns(Integer) }
        def c; end

        sig { params(c: Integer).void }
        def c=(c); end
```

lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,146 @@
+# typed: strict
+# frozen_string_literal: true
+
+module RBI
+  module TempHelpers
+    extend T::Sig
+  end
+
+  module Rewriters
+    class ReplaceAttributesWithMethods < Visitor
```
amomchilov, line 10: Name's a bit wordy, and "Replace" is kind of implied by "Rewriters::".

How about:

```suggestion
    class AttrToMethods < Visitor
```

lib/rbi/model.rb
```diff
@@ -42,6 +42,16 @@ def replace(node)
       self.parent_tree = nil
     end
 
+    sig { params(nodes: T::Enumerable[Node]).void }
+    def replace_with_multiple(nodes)
```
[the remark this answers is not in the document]
Morriar: It seems very niche. Let's keep it in the rewriter, we can promote it if we find another use case for it.

lib/rbi/model.rb
```diff
@@ -42,6 +42,16 @@ def replace(node)
...
+      tree = parent_tree
+      raise unless tree
+
+      # Does this work?
```
Morriar, line 50: Did you find a way to answer this question?

lib/rbi/model.rb
```diff
@@ -1153,6 +1163,43 @@ def initialize(
...
+        is_abstract: T::Boolean,
+        is_override: T::Boolean,
+        is_overridable: T::Boolean,
+        is_final: T::Boolean,
+        type_params: T::Array[String],
+        checked: T.nilable(Symbol),
+        loc: T.nilable(Loc),
+      ).returns(Sig)
+    end
+    def new_with(
```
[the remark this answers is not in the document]
Morriar: What's the benefits over calling `Sig.new` directly?

lib/rbi/model.rb
```diff
@@ -1171,6 +1218,15 @@ def ==(other)
         is_override == other.is_override && is_overridable == other.is_overridable && is_final == other.is_final &&
         type_params == other.type_params && checked == other.checked
     end
+
+    sig { override.returns(String) }
+    def inspect
```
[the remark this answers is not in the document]
Morriar: Let's not override the default `inspect`, you can call `node.string` for the same result

lib/rbi/rewriters/replace_attributes_with_methods.rb
Morriar: Yeah I like it more, other rewriters always use an action verb 👍

lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,146 @@
...
+      extend T::Sig
+      include TempHelpers
```
Morriar, line 12:
```suggestion

      include TempHelpers
```

lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,146 @@
...
+
+      sig { override.params(node: T.nilable(Node)).void }
+      def visit(node)
+        return unless node
```
Morriar, line 16:
```suggestion
        return unless node.is_a?(Tree)
```

so you can remove the case below

lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,146 @@
...
+        writers = attr.names.map do |name|
+          create_setter_method(name.to_s, sig, attribute_type, attr.visibility, attr.loc, attr.comments)
+        end
+
+        readers + writers
+      end
+
+      sig { params(attr: Attr).returns([T.nilable(Sig), T.nilable(String)]) }
+      def parse_sig_of(attr)
+        raise "Attributes cannot have more than 1 sig" if 1 < attr.sigs.count
```
Morriar, line 76: I would rather not have RBI do this check, as far as we know this is perfectly valid RBI syntax, the semantics of it should be left to Tapioca/Sorbet

lib/rbi/rewriters/replace_attributes_with_methods.rb
```diff
@@ -0,0 +1,146 @@
...
+            child.replace_with_multiple(new_methods)
+          end
+        end
+      end
+
+      private
+
+      sig { params(attr: Attr).returns(T::Array[Method]) }
+      def convert_to_methods(attr)
+        sig, attribute_type = parse_sig_of(attr)
```
Morriar, line 35: We could move this to where we actually create the resulting methods and let them `translate` the sig rather than passing it around.

* going from `attr_reader :a` to `def a; end` shouldn't require any sig translation
* going from `attr_writer :a` to `def a=(val); end` may require a sig translation
* going from `attr_accessor :a` requires a sig translation only for the writer

lib/rbi/model.rb
amomchilov: Yup, works :)

lib/rbi/rewriters/replace_attributes_with_methods.rb
amomchilov: Ah, these were indeed meant to be temporary while I refactored stuff around. They'll be gone shortly.

lib/rbi/rewriters/attr_to_methods.rb
```diff
@@ -0,0 +1,150 @@
...
+        case node
+        when Tree
+          node.nodes.dup.each do |child|
+            visit(child)
+            next unless (attr = child).is_a?(Attr)
+
+            new_methods = convert_to_methods(attr)
+
+            replace(child, with: new_methods)
+          end
```
Morriar, line 22: Can't we simplify this actually?

```suggestion
        return unless node

        case node
        when Attr
          new_methods = convert_to_methods(attr)
          replace(child, with: new_methods)
        when Tree
          visit_all(node.nodes)
        end
```

Morriar: You could even hoist up the logic from `convert_to_methods`?

lib/rbi/model.rb
```diff
@@ -380,6 +380,9 @@ def initialize(name, names, visibility: Public.new, sigs: [], loc: nil, comments
 
     sig { abstract.returns(T::Array[String]) }
     def fully_qualified_names; end
+
+    sig { abstract.returns(T::Array[Method]) }
+    def convert_to_methods; end
```
Morriar, line 385: Let's move this in lib/rbi/rewriters/attr_to_methods.rb with the other methods

lib/rbi/rewriters/attr_to_methods.rb
```diff
@@ -0,0 +1,151 @@
...
+module RBI
+  module Rewriters
+    class AttrToMethods < Visitor
+      extend T::Sig
+
+      sig { override.params(node: T.nilable(Node)).void }
+      def visit(node)
+        case node
+        when Tree
+          visit_all(node.nodes)
```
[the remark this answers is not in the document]
Morriar: I think you need a test iterating over multiple nodes:

```rb
class Foo
  def foo; end
  attr_accessor :bar
  def baz; end
end
```

lib/rbi/rewriters/attr_to_methods.rb
```diff
@@ -0,0 +1,151 @@
...
+
+        when Attr
+          new_methods = node.convert_to_methods
+          replace(node, with: new_methods)
```
Morriar, line 17:
```suggestion
          replace(node, with: node.convert_to_methods)
```

lib/rbi/rewriters/attr_to_methods.rb
```diff
@@ -0,0 +1,151 @@
...
+        end
+      end
+
+      private
+
+      sig { params(node: Node, with: T::Array[Node]).void }
+      def replace(node, with:)
+        tree = node.parent_tree
+        raise unless tree
```
Morriar, line 26: Let's raise a `AttrToMethods::Error`

lib/rbi/rewriters/attr_to_methods.rb
```diff
@@ -0,0 +1,151 @@
...
+      visitor.visit(self)
+    end
+  end
+
+  class Attr
+    private
+
+    sig(:final) { returns([T.nilable(Sig), T.nilable(String)]) }
+    def parse_sig
+      raise "Attributes cannot have more than 1 sig" if 1 < sigs.count
```
Morriar, line 49: Let's raise a `AttrToMethods::Error`

lib/rbi/rewriters/attr_to_methods.rb
```diff
@@ -0,0 +1,168 @@
...
+  module Rewriters
+    class AttrToMethods < Visitor
+      extend T::Sig
+
+      sig { override.params(node: T.nilable(Node)).void }
+      def visit(node)
+        case node
+        when Tree
+          visit_all(node.nodes.dup)
+
```
Morriar, line 29:
```suggestion
```

Morriar (approved):

PR #311, opened 2024-06-10
Add more specific error classes

Morriar:
We shouldn't raise bare errors as it forces the client to rescue `RuntimeErrror`, it's bad practice and bad design.

This PR adds specific error classes that can be rescued more cleanly.

This may be a breaking change since people rescuing `RuntimeError` to catch exceptions raised from `rbi` won't catch anything now since `RBI::Error` is a `StandardError` and will need to update their rescues.

### DIFF
--- a/lib/rbi.rb
+++ b/lib/rbi.rb
@@ -3,6 +3,11 @@
 
 require "sorbet-runtime"
 require "stringio"
+
+module RBI
+  class Error < StandardError; end
+end
+
 require "rbi/loc"
 require "rbi/model"
 require "rbi/visitor"
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -2,6 +2,8 @@
 # frozen_string_literal: true
 
 module RBI
+  class ReplaceNodeError < Error; end
+
   class Node
     extend T::Sig
     extend T::Helpers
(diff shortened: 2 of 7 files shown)


andyw8:
Feels like a bare `raise` (outside of a `rescue` block) is something that RuboCop should complain about, but I couldn't find a cop for it.

egiurleo (commented):
I feel like the names of the errors are inconsistent. Some of them are specific (e.g. `ReplaceNodeError`), and some are more general (e.g. `GroupNodes::Error`). I think it would be more consistent to give them all specific names. What do you think?

Morriar:
@egiurleo:

> I feel like the names of the errors are inconsistent. Some of them are specific (e.g. `ReplaceNodeError`), and some are more general (e.g. `GroupNodes::Error`). I think it would be more consistent to give them all specific names. What do you think?

Indeed, I like the `GroupNodes::Error` scheme as the error gets nested in the rewriter class it belongs to. The issue is with errors coming from the model, there is no namespace to define them into.

I could just use a base `Model::Error` if you prefer? But the `ReplaceNodeError` seemed more specific.

Another way would be to create a namespace for errors:
* `RBI::Errors::ReplaceNodeError`
* `RBI::Errors::UnknownGroupError`
* `RBI::Errors::UnknownNodeError`
* ...

egiurleo (approved):
Yeah I think this works too! It's very clear what each error is for.

PR #312, opened 2024-06-11
Remove redundant nil checks in `RBI::Visitor`

amomchilov:


lib/rbi/rewriters/remove_known_definitions.rb
```diff
@@ -81,8 +81,6 @@ def visit_all(nodes)
 
       sig { override.params(node: T.nilable(Node)).void }
       def visit(node)
-        return unless node
```
amomchilov, line 84: I'm not convinced removing these is a great idea. They're _technically_ redundant, but they also fail fast on `nil`, which would be a really common case to hit (rather than doing 3 `Class#===` checks below)

lib/rbi/rewriters/sort_nodes.rb
```diff
@@ -13,28 +13,28 @@ def visit(node)
...
-        original_order = node.nodes.map.with_index.to_h
+        original_order = node.nodes.each_with_index.to_h
 
         # The child nodes could contain private/protected markers. If so, they should not be moved in the file.
         # Otherwise, some methods could see their privacy change. To avoid that problem, divide the array of child
         # nodes into chunks based on whether any Visibility nodes appear, and sort the chunks independently. This
         # applies the ordering rules from the node_rank method as much as possible, while preserving visibility.
-        sorted_nodes = node.nodes.chunk do |n|
-          n.is_a?(Visibility)
-        end.flat_map do |_, nodes|
```
amomchilov, line 24: There's no Rubocop rule that can enforce this (last I checked), but having `end.something_else do ||` on the same line is kinda weird, and reads better when it's one operation per line IMO.

lib/rbi/rewriters/nest_non_public_methods.rb
```diff
@@ -8,33 +8,30 @@ class NestNonPublicMethods < Visitor
 
       sig { override.params(node: T.nilable(Node)).void }
       def visit(node)
-        return unless node
-
-        case node
-        when Tree
```
amomchilov, line 14: Removing these `case` statements with only one `when` worked out nicely.

PR #332, opened 2024-07-31
Add Type parsing

Morriar:
This PR adds parsing for the `Type` builder introduced in #331.

This allows the user to parse types from a string and get a `RBI::Type` from them:

```rb
type = Type.parse_string("T.nilable(Foo)")
type.class # => Type::Nilable
type.type # => Type::Simple("Foo")
```

Note that we do not migrate the `RBI::Parser` to parse types at this point, the general parser still uses Strings to represent types in signatures and T::Structs.

cc. @paracycle


lib/rbi/type_parser.rb
```diff
@@ -0,0 +1,320 @@
...
+      end
+
+      private
+
+      sig { params(node: T.any(Prism::ConstantReadNode, Prism::ConstantPathNode)).returns(Type) }
+      def parse_constant(node)
+        case node
+        when Prism::ConstantReadNode
+          # `Foo`
+          Type.simple(node.slice)
```
paracycle, line 56: I know you had concerns about the parser potentially raising when parsing non-sensical input instead of parsing it as is and allowing another system to show errors.

For that reason, I wonder if we should use the constructors directly here, instead of using the helper methods.

Morriar: I'm still conflicted on the choice with made regarding the builder API.

Calling `new` directly is a good escape hatch but by making it private we made it impossible to use for external users. Maybe the right thing to do would be to keep the `new` public?

In the meantime you're right, I changed all the calls from the parser to use `new` directly rather than the builder interface.

paracycle: Agreed, but it is easier to make the `new` public later than to make it private later. So, I think I like how we are exposing a more minimal API for the time being.

PR #333, opened 2024-07-31
Fix comments parsing for RBI::Method

Morriar:
Before this PR, parsing a file such as this one:

```rb
# some comment
def foo(x); end
```

End up associating the comment to the `x` parameter:

```rb
def foo(
  x, # some comment
); end
```

This PR fixes the way we associate the comment to the method so the result is accurate:

```rb
# some comment
def foo(x); end
```


lib/rbi/parser.rb
```diff
@@ -242,10 +242,10 @@ def visit_def_node(node)
         @last_node = node
         current_scope << Method.new(
           node.name.to_s,
+          comments: current_sigs_comments + node_comments(node),
```
egiurleo, line 245: I'm confused why moving this kwarg would make a difference? But maybe I'm missing something?

paracycle (approved):

PR #342, opened 2024-08-05
Fix parsing of `T::Enum`

Morriar:
The parser was returning a `Class` node for subclasses of `T::Enum` when we have a more precise type of node with `TEnum`.

This also showed that we were not visiting the `TEnum` properly as we called the visit on it as if a `Class`.

Maybe `T::Enum` shouldn't descend from `Class`?

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -175,12 +175,22 @@ module RBI
       sig { override.params(node: Prism::ClassNode).void }
       def visit_class_node(node)
         @last_node = node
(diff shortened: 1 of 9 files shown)


egiurleo (approved):
It is somewhat weird that T::Enum descendants are classes when they don't behave like classes and shouldn't be used like classes.

Morriar:
I'm going to merge the parsing fix, the API can be changed in another PR as this is mostly going to be a breaking change 👍

PR #349, opened 2024-08-06
Make more tests using parsing for their input

Morriar:
While a bit wasteful, writing the test input as a real RBI string rather than crafting in manually using the RBI classes is much easier to write/read.

Especially around nested scopes:

```rb
tree = parse_rbi(<<~RBI)
  class Foo
    class Bar
      class Baz; end
    end
  end
RBI
```

makes it easier to understand the scope hierarchy than:

```rb
tree = Tree.new
scope1 = Class.new("Foo")
scope2 = Class.new("Bar")
scope2 << Class.new("Baz")
scope1 << scope2
tree << scope1
```


egiurleo (approved):
This is waaaaay better

PR #354, opened 2024-08-06
Make TEnumBlock behave as a scope

Morriar:
To represent `T::Enum` subclasses like this one:

```rb
class MyEnum < T::Enum
  enums do
    # Some comments
    A = new
    B = new
  end
end
```

We currently use a special node called `TEnumBlock` that only accepts an array of names representing the constants like `A`, `B`. On top of being unnecessarily restrictive and complex, it doesn't allow storing the comments attached to each enum constant.

In reality, `enums do .. end` is nothing more than a block that can accept constants but also other types of nodes. This PR changes the superclass of `TEnumBlock` to be a `Tree` that can contain any kind of nodes including `Const` to represent the `A`, `B` etc. This means we can now attach comments to the constants and parse the TEnum properly.

This is a breaking change as we change the API to build `TEnum` which may break some clients.


lib/rbi/parser.rb
```diff
@@ -383,25 +383,11 @@ def visit_call_node(node)
...
-          current_scope << TEnumBlock.new(
-            body.body.map { |stmt| T.cast(stmt, Prism::ConstantWriteNode).name.to_s },
-            loc: node_loc(node),
-            comments: node_comments(node),
-          )
+          scope = TEnumBlock.new(loc: node_loc(node), comments: node_comments(node))
+          current_scope << scope
+          @scopes_stack << scope
+          visit(node.block)
+          @scopes_stack.pop
```
paracycle, line 390: Shouldn't we be checking that the `enums` call has actually been supplied a block **and** no arguments?

paracycle (commented):
The changes look fine to me, but I can't understand how one is supposed to extract all enum members from the resulting `TEnumBlock` object? Do you need to find all the constants defined in that scope with the value `new`?

Morriar:
> The changes look fine to me, but I can't understand how one is supposed to extract all enum members from the resulting `TEnumBlock` object? Do you need to find all the constants defined in that scope with the value `new`?

Good point, I added a `TEnum#members` method to make gathering the enum constants super easily:

```rb
enum = TEnum.new("Enum") do |enum|
  enum << TEnumBlock.new do |block|
    block << Const.new("A", "new")
    block << Const.new("B", "new")
  end
end

assert_equal(["A", "B"], enum.members)
```

lib/rbi/parser.rb
```diff
@@ -383,11 +383,20 @@ def visit_call_node(node)
...
+            current_scope << scope
+            @scopes_stack << scope
+            visit(node.block)
+            @scopes_stack.pop
+          else
+            current_scope << Send.new(
+              message,
+              parse_send_args(node.arguments),
+              loc: node_loc(node),
+              comments: node_comments(node),
```
egiurleo, line 397: Just to make sure I'm understanding -- if arguments are passed in, we treat `enums` as a regular method call? (Which makes sense to me)

egiurleo (approved):

lib/rbi/model.rb
```diff
@@ -1341,41 +1341,43 @@ def initialize(name, loc: nil, comments: [], &block)
...
+      "#{parent_scope&.fully_qualified_name}.enums"
     end
 
-    sig { params(name: String).void }
-    def <<(name)
-      @names << name
+    # Find the names defined by the enum.
+    sig { returns(T::Array[String]) }
+    def members
+      nodes.grep(Const).select { |const| const.value == "new" }.map(&:name)
```
paracycle, line 1375: What about cases where we might have:
```ruby
class Foo < T::Enum
  enums do
    A = new("A")
  end
end
```
Is this not a concern for RBI files, I wonder?

paracycle (approved):

PR #361, opened 2024-08-28
Bump Prism requirement to v1

vinistock:
Bump our Prism requirement to v1, so that we can upgrade other projects too.


rbi.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
     "Rakefile",
   ]
 
-  spec.add_dependency("prism", ">= 0.18.0", "< 1.0.0")
+  spec.add_dependency("prism", ">= 1.0.0", "< 2.0.0")
```
rafaelfranca, line 26:
```suggestion
  spec.add_dependency("prism", "~> 1.0")
```

PR #368, opened 2024-10-03
Do not rely on return type for `attr_writer` RBS signatures

Morriar:
The current RBS translation mechanism for attr_writers relies on the return type of its signature. However, the return type can't always be trusted as both these syntax are valid for Sorbet:

```rb
sig { params(x: Integer).returns(Integer) }
attr_writer :x

sig { params(y: Integer).void }
attr_writer :y
```

Instead, we should rely on the first parameter which we can be sure to find since Sorbet requires it.


amomchilov (approved):
Subtle, good catch

paracycle (approved):
Makes sense!

PR #384, opened 2024-12-10
Translate RBI `attached_class` to RBS `instance`

Morriar:
The `attached_class` in RBI:

```rb
sig { returns(T.attached_class) }
def foo; end
```

should be translated to `instance` in RBS:

```rbs
def foo: -> instance
```

### DIFF
--- a/lib/rbi/rbs_printer.rb
+++ b/lib/rbi/rbs_printer.rb
@@ -880,7 +880,7 @@ module RBI
 
     sig { params(type: Type::AttachedClass).void }
     def visit_attached_class(type)
-      @string << "attached_class"
+      @string << "instance"
     end
 
     sig { params(type: Type::Nilable).void }
--- a/test/rbi/rbs_printer_test.rb
+++ b/test/rbi/rbs_printer_test.rb
@@ -519,6 +519,17 @@ module RBI
       assert_empty(rbi.rbs_string)
     end
 
+    def test_print_attached_class
+      rbi = parse_rbi(<<~RBI)
+        sig { returns(T.attached_class) }
+        def foo; end
+      RBI
+
+      assert_equal(<<~RBI, rbi.rbs_string)
+        def foo: -> instance
+      RBI
+    end
+
     def test_print_t_structs
       rbi = parse_rbi(<<~RBI)
         class Foo < T::Struct; end


vinistock (approved):

paracycle:
Are you sure that that is the right translation? The [example](https://github.com/ruby/rbs/blob/master/docs/syntax.md#examples-of-classish-context) in the RBS docs doesn't make it sound like it is so:
```ruby
class Foo
  attr_reader parent: class

  def foo: () -> instance

  @@foos: Array[instances]

  include Enumerable[class]
end
```
Note that the `foo` method is an instance method, so the concept AttachedClass does not even make sense.

Does `instance` mean a more general "attached object" then?

Morriar:
I'm not sure about the `foo` method in this case, the class attribute would make sense (it seems there is [a typo](https://github.com/ruby/rbs/pull/2139) in the example though):

```
class Foo
  attr_reader parent: class
  @@foos: Array[instances]
end
```

The document [says](https://github.com/ruby/rbs/blob/master/docs/syntax.md#base-types): 
> `instance` denotes the type of instance of the class. `class` is the singleton of the class.

Looking at [how it is used](https://github.com/ruby/rbs/blob/master/core/rubygems/requirement.rbs#L50) in the RBS files for core it seems to mean `T.attached_class`?

The usage in `foo` is indeed surprising. I though that depending on the context it could mean `T.self_type` but there is already [`self`](https://github.com/ruby/rbs/blob/master/docs/syntax.md#base-types) in RBS. I'll inquire.

amomchilov (approved):
LGTM, pending clarification of how `instance` is intended to work in RBS.

Morriar:
> I though that depending on the context it could mean `T.self_type` but there is already [`self`](https://github.com/ruby/rbs/blob/master/docs/syntax.md#base-types) in RBS. I'll inquire.

So it's indeed [dependent on the context](https://github.com/ruby/rbs/pull/2139/files/ec12054aceb3a50984df29654022b7254f90ddcc#r1885461234).

* On a class method it does return the type of the instance, so the same as `T.attached_class`.
* On an instance method it behaves as `self`, so it should be translated as `T.self_type`.

I think this is useful for RBS methods declared with the syntax `def self?.foo: instance` where the return type will be `instance` in the case of the singleton and `self` in the case of the instance.

Since we come from RBI, if we find a `T.attached_class` we can presume that it was used in the context of a singleton method (or Sorbet would have [errored](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20returns%28T.attached_class%29%20%7D%0A%20%20def%20foo%0A%20%20end%0A%0A%20%20sig%20%7B%20returns%28T.attached_class%29%20%7D%0A%20%20def%20self.bar%0A%20%20%20%20new%0A%20%20end%0A%0A%20%20class%20%3C%3C%20self%0A%20%20%20%20extend%20T%3A%3ASig%0A%0A%20%20%20%20sig%20%7B%20returns%28T.attached_class%29%20%7D%0A%20%20%20%20def%20self.baz%0A%20%20%20%20%20%20new%0A%20%20%20%20end%0A%20%20end%0Aend)) so I think it's safe to translate it to `instance`.

PR #393, opened 2025-01-14
Handle spaces in types during RBS translation

Morriar:
I found some signatures formatted as follow:

```rb
sig { returns(Foo::
              Bar) }
```

While this is valid in Ruby/RBI, having spaces or new lines in a type is a syntax error in RBS.

### DIFF
--- a/test/rbi/parser_test.rb
+++ b/test/rbi/parser_test.rb
@@ -88,6 +88,16 @@ module RBI
       assert_equal(rbi, tree.string)
     end
 
+    def test_parse_constants_with_newlines
+      rbi = <<~RBI
+        sig { returns(Foo::
+              Bar) }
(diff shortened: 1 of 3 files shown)


egiurleo (approved):

test/rbi/parser_test.rb
```diff
@@ -88,6 +88,17 @@ def test_parse_constants
...
     end
 
+    def test_parse_constants_with_newlines
+      rbi = <<~RBI
+        sig { returns(Foo::
+              Bar) }
+      RBI
+
+      tree = parse_rbi(rbi)
+      puts tree.string
```
vinistock, line 98: ✂️

vinistock (approved):

PR #394, opened 2025-01-14
Fix RBS translation of non-block type in block position

Morriar:
Consider this RBI signature:

```rb
sig { params(block: T.proc.void).void }
def foo(&block); end
```

The RBS equivalent to define the block type is this:

```rbs
def foo: { () -> void } -> void
```

Now on the tricky part, RBS doesn't support anything else than an actual block type definition in block position so this would be invalid syntax:

```rbs
def foo: { FOO } -> void
```

This becomes an issue when using RBI type aliases such as this:

```rb
FOO = T.type_alias { T.proc.void }

sig { params(block: FOO) }
def foo(&block); end
```

This could be changed upstream following an RFC against RBS but for now let's just ignore the case and generate the RBS as untyped:

```rbs
FOO: untyped

def foo: { (?) -> untyped } -> void
```


egiurleo (approved):
Nice. I like your prioritization here!

vinistock (approved):
Can't we grab the contents of the type alias and put that as the type?

Morriar:
> Can't we grab the contents of the type alias and put that as the type?

The issue is that the alias can exist anywhere in the codebase include in `.rbi` files. We would need a global knowledge to be able to do this.

We could imagine doing this at Sorbet level or make a smarter translator that uses the code index but I only found a couple of sigs using this pattern so that's a lot of complexity for relatively low benefits.

Maybe a simpler approach would be to write a cop to spot those cases and fix them manually before running the translator.

PR #399, opened 2025-01-14
Fix RBS translation of attr_writer signatures

Morriar:
I found some occurrences where the RBI signature for attribute writers where using a different type for the parameter and the return:

```rb
sig { params(baz: String).returns(T.nilable(String)) }
attr_writer :x
```

In this case Sorbet would only consider the return type so we need to keep this information during the translation to RBS:

```rbs
attr_writer x: String?
```

### DIFF
--- a/lib/rbi/rbs_printer.rb
+++ b/lib/rbi/rbs_printer.rb
@@ -273,19 +273,24 @@ module RBI
 
     sig { params(node: RBI::Attr, sig: Sig).void }
     def print_attr_sig(node, sig)
+      ret_type = sig.return_type
+
       type = case node
-      when AttrAccessor, AttrReader
-        parse_type(sig.return_type)
+      when AttrReader, AttrAccessor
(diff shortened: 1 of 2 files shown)


lib/rbi/rbs_printer.rb
```diff
@@ -273,19 +273,21 @@ def visit_attr(node)
...
       else
-        first_arg = sig.params.first
-        if first_arg
-          parse_type(first_arg.type)
+        arg_type = sig.params.first
+        if arg_type && (ret_type.is_a?(Type::Void) || ret_type == "void")
+          parse_type(arg_type.type).rbs_string
         else
-          Type.untyped
+          parse_type(ret_type).rbs_string
```
egiurleo, line 286: Not blocking, but I think it would be helpful to have a comment here that in the case of mismatching arg and return types, Sorbet will default to the return type

egiurleo (approved):

PR #420, opened 2025-02-25
Introduce `RBSSig` node

KaanOzkan:
A simple node that's currently used for printing RBS signatures correctly.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -423,13 +423,17 @@ module RBI
     #: Array[Sig]
     attr_accessor :sigs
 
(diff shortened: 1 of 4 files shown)


lib/rbi/model.rb
```diff
@@ -982,6 +987,21 @@ def ==(other)
     end
   end
 
+  # RBS signatures
+
+  class RBSSig < NodeWithComments
```
Morriar, line 992: I'm not a fan of the name, what about `CommentSig` or `SigComment`?

lib/rbi/printer.rb
```diff
@@ -466,6 +469,15 @@ def visit_sig(node)
       end
     end
 
+    # @override
+    #: (RBSSig node) -> void
+    def visit_rbs_sig(node)
+      print_loc(node)
+      printt("#")
```
Morriar, line 476: Let's strip the `:` when we create the sig so it's nicer to handle, then we add it back when we print:

```suggestion
      printt("#:")
```

Morriar (approved):

PR #423, opened 2025-03-04
Introduce RBSComment node

Morriar:
Parse and storer RBS comments on nodes.

For now RBS comments are just stored as a special kind of comments, we'll then plug this with #420 to parse them into actual RBS signatures.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -82,6 +82,18 @@ module RBI
     end
   end
 
+  # A comment representing a RBS type prefixed with `#:`
+  class RBSComment < Comment
+    extend T::Sig
+
+    #: (Object other) -> bool
+    def ==(other)
+      return false unless other.is_a?(RBSComment)
+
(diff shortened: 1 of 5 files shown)


lib/rbi/model.rb
```diff
@@ -82,6 +82,18 @@ def initialize(loc: nil)
     end
   end
 
+  # A comment representing a RBS signature prefixed with `#:`
+  class RBSComment < Comment
```
vinistock, line 86: The comment here says this class represents an RBS signature. However, the name is `RBSComment`. Given that we'll also have the concept of inline type assertion in comments, should we instead name this `RBSSignature`?

Or if this is intended to be used by any RBS comment, then we can change the comment on top of the class.

vinistock (approved):

lib/rbi/model.rb
Morriar: Changed for `RBS type`

PR #424, opened 2025-03-04
RBS sig rewriter

Morriar:
Add all the tooling necessary to transform a file containing `#: ` RBS signature comments into a file containing Sorbet sigs instead:

```rb
rbi = <<~RBI
  #: -> void
  def foo; end

  #: Integer
  attr_writer :bar
RBI

tree = parse_rbi(rbi)
tree.translate_rbs_sigs!
puts tree.string
```

will output:

```rb
sig { void }
def foo; end

sig { params(bar: Integer).returns(Integer) }
attr_writer :bar
```


lib/rbi/rbs/type_translator.rb
```diff
@@ -0,0 +1,164 @@
...
+          namespace = name.namespace
+          string << "::" if namespace.absolute?
+
+          path = namespace.path
+          string << path.join("::")
+          string << "::" if path.any?
+
+          string << name.name.to_s
+
+          string
```
paracycle, line 115: I have a few points about this:
1. Let's not use `String.new` for creating a mutable string, since it has a footgun: It always created an `ASCII-8BIT` string, despite the encoding of the Ruby file. Instead, we should use `"".dup` or `+""` to create mutable strings, and
2. I believe this is exactly what `RBS::TypeName#to_s` does internally anyway, so I am not sure why we need this at all. Ref: https://github.com/ruby/rbs/blob/master/lib/rbs/type_name.rb#L35-L37 and https://github.com/ruby/rbs/blob/master/lib/rbs/namespace.rb#L75-L82

lib/rbi/rbs/type_translator.rb
```diff
@@ -0,0 +1,164 @@
...
+        end
+
+        #: (::RBS::Types::Function) -> Type
+        def translate_function(type)
+          proc = Type.proc
+
+          index = 0
+
+          type.required_positionals.each do |param|
+            proc.proc_params[:"arg#{index}"] = translate(param.type)
```
paracycle, line 125: These params could have declared names, no? Shouldn't we be using anonymous names only when the names don't exist?

lib/rbi/rewriters/translate_rbs_sigs.rb
```diff
@@ -0,0 +1,91 @@
...
+        when AttrAccessor, AttrReader, AttrWriter
+          rbs_comments = extract_rbs_comments(node)
+          rbs_comments.each do |comment|
+            node.sigs << translate_rbs_attr_type(node, comment)
+          end
+        when Method
+          rbs_comments = extract_rbs_comments(node)
+          rbs_comments.each do |comment|
+            node.sigs << translate_rbs_method_type(node, comment)
+          end
```
paracycle, line 29: Nitpick: I think this would read better as:
```suggestion
          transform_rbs_comments(node) do |comment|
            translate_rbs_method_type(node, comment)
          end
```
if we renamed `extract_rbs_comments` to `transform_rbs_comments` and allowed it to accept a block that would get `yield`ed to inside.

lib/rbi/rewriters/translate_rbs_sigs.rb
```diff
@@ -0,0 +1,91 @@
...
+          case comment
+          when RBSComment
+            rbs_sigs << comment
+          else
+            node.comments << comment
+          end
+        end
+
+        rbs_sigs
+      end
```
paracycle, line 52: In line with my suggestion above, this could be:
```suggestion
      def transform_rbs_comments(node, &blk)
        comments = node.comments.dup
        node.comments.clear

        comments.each do |comment|
          case comment
          when RBSComment
            node.sigs << blk.call(comment)
          else
            node.comments << comment
          end
        end
      end
```

lib/rbi/rewriters/translate_rbs_sigs.rb
Morriar: https://github.com/Shopify/rbi/pull/424#discussion_r1980156335

lib/rbi/rbs/type_translator.rb
Morriar: Oh yeah, we should keep the names if they exist. Changed the code and added a test 👍

paracycle (approved):

lib/rbi/rbs/method_type_translator.rb
```diff
@@ -0,0 +1,125 @@
...
+          index += 1
+        end
+
+        rest_keyword = type.rest_keywords
+        if rest_keyword
+          result.params << translate_function_param(rest_keyword, index)
+        end
+
+        return_type = translate_type(type.return_type)
+        result.return_type = return_type
```
vinistock, line 101: Nit: we don't seem to need this variable.
```suggestion
        result.return_type = translate_type(type.return_type)
```

PR #430, opened 2025-03-06
Fix RBI -> RBS translation for generics

Morriar:
Two fixes in there:

1. Properly translate generic classes so RBS `Foo[A, B]` is not just translated as `Foo`
2. Translate generic `Array` to `T::Array` and friends

### DIFF
--- a/lib/rbi/rbs/type_translator.rb
+++ b/lib/rbi/rbs/type_translator.rb
@@ -99,7 +99,8 @@ module RBI
         def translate_class_instance(type)
           return Type.simple(type.name.to_s) if type.args.empty?
 
-          T.unsafe(Type).generic(type.name.to_s, *type.args.map { |arg| translate(arg) })
+          type_name = translate_t_generic_type(type.name.to_s)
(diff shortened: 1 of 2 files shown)


lib/rbi/rbs/type_translator.rb
```diff
@@ -141,6 +149,28 @@ def translate_function(type)
...
+            "T::Enumerator"
+          when "Hash"
+            "T::Hash"
+          when "Set"
+            "T::Set"
+          when "Range"
+            "T::Range"
+          else
+            type_name
+          end
```
paracycle, line 172: I think we are missing `T::Enumerator::Chain` and `T::Enumerator::Lazy` here: https://github.com/sorbet/sorbet/blob/master/rbi/sorbet/t.rbi#L323-L376

paracycle: Also, what if the name was fully qualified, like `::Array[String]`?

KaanOzkan (commented):
Able to generate correct sigs for ruby-lsp with this branch 👍

KaanOzkan (approved):

PR #436, opened 2025-03-13
Do not parse rdoc annotations as RBS comments

Morriar:
So this:

```rb
#:nodoc:
#:yields: something
def foo; end
```

is not considered having RBS comments

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -576,7 +576,9 @@ module RBI
       def parse_comment(node)
         loc = Loc.from_prism(@file, node.location)
         string = node.location.slice
-        if string.start_with?("#:")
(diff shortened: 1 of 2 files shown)


lib/rbi/parser.rb
```diff
@@ -576,7 +576,8 @@ def node_comments(node)
       def parse_comment(node)
         loc = Loc.from_prism(@file, node.location)
         string = node.location.slice
-        if string.start_with?("#:")
+        # We also ignore RDoc directives such as `:nodoc:`. See https://docs.ruby-lang.org/en/2.1.0/RDoc/Markup.html.
```
paracycle, line 579: Sorry, I realized that I linked to a really old version of the doc and updated my comment. Do you mind updating the link to that as well?

PR #441, opened 2025-03-26
Parse T::Enum values as special nodes

Morriar:
Apply idea from https://github.com/Shopify/rbi/pull/408#issuecomment-2755037767.

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -228,6 +228,17 @@ module RBI
 
         current_scope << if struct
           struct
+        elsif t_enum_value?(node)
+          TEnumValue.new(
+            case node
+            when Prism::ConstantWriteNode
+              node.name.to_s
+            when Prism::ConstantPathWriteNode
+              node_string!(node.target)
+            end,
+            loc: node_loc(node),
+            comments: node_comments(node),
+          )
         else
           adjusted_node_location = adjust_prism_location_for_heredoc(node)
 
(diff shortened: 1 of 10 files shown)


lib/rbi/parser.rb
```diff
@@ -840,6 +851,11 @@ def set_root_tree_loc
       def type_variable_definition?(node)
         node.is_a?(Prism::CallNode) && (node.message == "type_member" || node.message == "type_template")
       end
+
+      #: (Prism::Node? node) -> bool
+      def t_enum_value?(node)
+        current_scope.is_a?(TEnumBlock) && node.is_a?(Prism::ConstantWriteNode) && node.value.slice == "new"
```
paracycle, line 857: I believe the node value could be more than `new`. For example, it can be `new("Foo")`. Does this handle that case?

lib/rbi/printer.rb
```diff
@@ -542,6 +542,16 @@ def visit_tenum_block(node)
...
     end
 
+    # @override
+    #: (TEnumValue node) -> void
+    def visit_tenum_value(node)
+      print_blank_line_before(node)
+      print_loc(node)
+      visit_all(node.comments)
+
+      printl("#{node.name} = new")
```
paracycle, line 552: Again, we might be missing the serialized value of the enum here.

paracycle (commented):
I like the changes, but I think we are missing some edge-case handling.

lib/rbi/parser.rb
Morriar: Ah you're right, we need to parse it but we don't care about storing it 👍

Morriar: Done 👍

lib/rbi/printer.rb
Morriar: Added in `parser_test` instead 👍

paracycle (approved):

PR #444, opened 2025-03-28
Check shims duplicate in CI

Morriar:
I found a duplicate that should have been caught on CI.


.github/workflows/main.yml
```diff
@@ -29,3 +29,6 @@ jobs:
 
       - name: Check exported RBI files
         run: bin/export-rbi --check-sync
+
+      - name: Check shims
+        run bin/tapioca check-shims
```
KaanOzkan, line 34:
```suggestion
        run: bin/tapioca check-shims
```

PR #445, opened 2025-03-28
Const holds a type rather than a value

Morriar:
Follow up on https://github.com/Shopify/rbi/pull/408#issuecomment-2755037767.

In the context of an RBI file, does it really make sense to hold the value of constants?

This PR proposes to change `Const` so it only stores the type we find a the `T.let` if any.

So this:

```rb
A = 42
B = T.let(42, Integer)
C = foo(T.let(42, Integer))
```

gives this:

```rb
A = T.let(T.unsafe(nil), T.untyped)
B = T.let(T.unsafe(nil), Integer)
C = T.let(T.unsafe(nil), T.untyped)
```

This is quite a breaking change for the gem's API and should be released with at least a minor version bump.

I think we can apply the same logic to optional positional and keyword parameters 🤔

### DIFF
--- a/test/rbi/parser_test.rb
+++ b/test/rbi/parser_test.rb
@@ -75,17 +75,26 @@ module RBI
     def test_parse_constants
       rbi = <<~RBI
         Foo = 42
-        Bar = "foo"
+        Bar = T.let("foo", String)
         Baz = Bar
         A = nil
         B = :s
-        C = T.nilable(String)
-        D = A::B::C
+        C = T.let(nil, T.nilable(String))
+        D = T.let(A::B::C, T.class_of(A::B::C))
         A::B::C = Foo
       RBI
 
       tree = parse_rbi(rbi)
-      assert_equal(rbi, tree.string)
+      assert_equal(<<~RBI, tree.string)
+        Foo = T.let(T.unsafe(nil), T.untyped)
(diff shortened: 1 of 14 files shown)


test/rbi/parser_test.rb
```diff
@@ -75,17 +75,26 @@ def m2; end
...
-        D = A::B::C
+        C = T.let(nil, T.nilable(String))
+        D = T.let(A::B::C, T.class_of(A::B::C))
         A::B::C = Foo
       RBI
 
       tree = parse_rbi(rbi)
-      assert_equal(rbi, tree.string)
+      assert_equal(<<~RBI, tree.string)
+        Foo = T.let(T.unsafe(nil), T.untyped)
```
KaanOzkan, line 89: Do we consider RBI files as documentation? Could this be a confusing change?

st0012: Also, will this create huge RBI changes when it's adopted by Tapioca?

st0012:
Given it's a breaking change, can the description also include the benefits it will bring too?

paracycle (approved):
I don't think this is a breaking change at all. The fact that we were including actual constant values in RBI files wasn't intentional and there should be nothing that relies on then being there.

PR #456, opened 2025-05-08
Add support for `T::Sig::WithoutRuntime`

Morriar:
This PR introduces a few changes:

* Signatures can be marked as `without_runtime`
* Signature without_runtime are printed as `T::Sig::WithoutRuntime.sig`
* Parse signature with `T::Sig::WithoutRuntime.sig`
* Handle RBS <> RBI translate for `T::Sig::WithoutRuntime.sig`

I also spotted an error in the parser where random calls to `sig` like `Foo.sig` would be parsed as actual `Sig` instances while it shouldn't. Fixed those.

### DIFF
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -1153,6 +1153,12 @@ class RBI::Parser::Visitor < ::Prism::Visitor
 
   sig { params(node: ::Prism::Node).returns(::String) }
   def node_string!(node); end
+
+  sig { params(node: T.nilable(::Prism::Node)).returns(T::Boolean) }
+  def self?(node); end
+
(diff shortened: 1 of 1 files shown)


test/rbi/rbs_printer_test.rb
```diff
@@ -398,6 +401,10 @@ def foo: -> void
 
         # @override(allow_incompatible: true)
         def bar: -> String?
+
+        # @without_runtime
```
vinistock, line 405: Why do we want to remember if a signature had originally been marked as without runtime? To translate back?

vinistock (approved):

egiurleo (approved):
🥳

paracycle (approved):

PR #461, opened 2025-05-16
Handle multiline RBS signatures

Morriar:
So this is parsed properly:

```rb
#: (
#|   Integer
#| ) -> void
def foo(x); end
```

### DIFF
--- a/lib/rbi/loc.rb
+++ b/lib/rbi/loc.rb
@@ -31,6 +31,17 @@ module RBI
       @end_column = end_column
     end
 
+    #: (Loc) -> Loc
+    def join(other)
+      Loc.new(
+        file: file,
+        begin_line: begin_line,
+        begin_column: begin_column,
+        end_line: other.end_line,
+        end_column: other.end_column,
+      )
+    end
+
     #: -> String
     def to_s
       if end_line && end_column
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -607,28 +607,66 @@ module RBI
         start_line = node.location.start_line
         start_line -= 1 unless @comments_by_line.key?(start_line)
 
+        rbs_continuation = [] #: Array[Prism::Comment]
+
         start_line.downto(1) do |line|
           comment = @comments_by_line[line]
           break unless comment
 
-          comments.unshift(parse_comment(comment))
+          text = comment.location.slice
+
+          # If we find a RBS comment continuation `#|`, we store it until we find the start with `#:`
+          if text.start_with?("#|")
+            rbs_continuation << comment
+            @comments_by_line.delete(line)
+            next
+          end
+
+          loc = Loc.from_prism(@file, comment.location)
+
+          # If we find the start of a RBS comment, we create a new RBSComment
+          # Note that we ignore RDoc directives such as `:nodoc:`
+          # See https://ruby.github.io/rdoc/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Directives
+          if text.start_with?("#:") && !(text =~ /^#:[a-z_]+:/)
+            text = text.sub(/^#: ?/, "").rstrip
+
+            # If we found continuation comments, we merge them in reverse order (since we go from bottom to top)
+            rbs_continuation.reverse_each do |rbs_comment|
+              continuation_text = rbs_comment.location.slice.sub(/^#\| ?/, "").strip
+              continuation_loc = Loc.from_prism(@file, rbs_comment.location)
+              loc = loc.join(continuation_loc)
+              text = "#{text}#{continuation_text}"
+            end
+
+            rbs_continuation.clear
+            comments.unshift(RBSComment.new(text, loc: loc))
+          else
+            # If we have unused continuation comments, we should inject them back to not lose them
+            rbs_continuation.each do |rbs_comment|
+              comments.unshift(parse_comment(rbs_comment))
+            end
+
+            rbs_continuation.clear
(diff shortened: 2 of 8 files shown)


lib/rbi/loc.rb
```diff
@@ -31,6 +31,17 @@ def initialize(file: nil, begin_line: nil, end_line: nil, begin_column: nil, end
...
+    #: (Loc) -> Loc
+    def join(other)
+      Loc.new(
+        file: file,
+        begin_line: begin_line,
+        begin_column: begin_column,
+        end_line: other.end_line,
+        end_column: other.end_column,
+      )
+    end
```
paracycle, line 43: Isn't this assuming that `other` will be further up in the document than `self`? What if I pass an `other` with `end_line` that is smaller than `self.begin_line`?

This assumption works for your current use-case, but, IMO, isn't correct for a general purpose method like `join`, especially when the method doesn't do any bounds check to reject invalid values to assert the assumption.

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
+            # If we found continuation comments, we merge them in reverse order (since we go from bottom to top)
+            rbs_continuation.reverse_each do |rbs_comment|
+              continuation_text = rbs_comment.location.slice.gsub(/^#\| ?/, "").strip
+              text = "#{text}#{continuation_text}"
+              loc = loc.join(Loc.from_prism(@file, rbs_comment.location))
+            end
+
+            comments.unshift(RBSComment.new(text, loc: loc))
+          else
+            # If we have unused continuation comments, we should injext them back to not lose them
```
paracycle, line 642:
```suggestion
            # If we have unused continuation comments, we should inject them back to not lose them
```

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
+            end
+
+            comments.unshift(parse_comment(comment))
+          end
+
+          rbs_continuation.clear
           @comments_by_line.delete(line)
         end
 
+        # If we have unused continuation comments, we should injext them back to not lose them
```
paracycle, line 654:
```suggestion
        # If we have unused continuation comments, we should inject them back to not lose them
```

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
+            next
+          end
+
+          loc = Loc.from_prism(@file, node.location)
+
+          # If we find the start of a RBS comment, we create a new RBSComment
+          # Note that we ignore RDoc directives such as `:nodoc:`
+          # See https://ruby.github.io/rdoc/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Directives
+          if text.start_with?("#:") && !(text =~ /^#:[a-z_]+:/)
+            text = text.gsub(/^#: ?/, "").rstrip
```
st0012, line 631: Nit: Since the text should only has one `^#:`, we can use `sub` instead.

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
+
+            # If we found continuation comments, we merge them in reverse order (since we go from bottom to top)
+            rbs_continuation.reverse_each do |rbs_comment|
+              continuation_text = rbs_comment.location.slice.gsub(/^#\| ?/, "").strip
```
st0012, line 635: Same as the above.

lib/rbi/parser.rb
st0012, line 642:
```suggestion
            # If we have unused continuation comments, we should inject them back to not lose them
```

lib/rbi/parser.rb
st0012, line 654:
```suggestion
        # If we have unused continuation comments, we should inject them back to not lose them
```

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
+              text = "#{text}#{continuation_text}"
+              loc = loc.join(Loc.from_prism(@file, rbs_comment.location))
+            end
+
+            comments.unshift(RBSComment.new(text, loc: loc))
```
st0012, line 640: Should we clear `rbs_continuation` after this?

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
+        rbs_continuation.each do |rbs_comment|
+          comments.unshift(parse_comment(rbs_comment))
+        end
+
         comments
       end
 
       #: (Prism::Comment node) -> Comment
       def parse_comment(node)
+        text = node.location.slice.gsub(/^# ?/, "").rstrip
```
st0012, line 664: Can be `sub` instead

lib/rbi/parser.rb
Morriar: It's being cleared line 650

lib/rbi/parser.rb
```diff
@@ -607,28 +607,63 @@ def node_comments(node)
...
         start_line.downto(1) do |line|
           comment = @comments_by_line[line]
           break unless comment
 
-          comments.unshift(parse_comment(comment))
+          text = comment.location.slice
+
+          # If we find a RBS comment continuation `#|`, we store it until we find the start with `#:`
+          if text.start_with?("#|")
+            rbs_continuation << comment
```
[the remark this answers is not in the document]
Morriar: Indeed, we could.

PR #479, opened 2025-06-05
Normalize and simplify types before printing them as RBS

Morriar:
This will avoid some translation bug where nested nilables such as `T.nilable(T.nilable(X))` get incorrectly translated as `X??`.

This PR introduce two important concept:

1. Type normalization: transforms a type into a canonical version that is easy to analyze.
  
   ```rb
    # Returns a normalized version of the type.
    #
    # Normalized types are meant to be easier to process, not to read.
    # For example, `T.any(TrueClass, FalseClass)` instead of `T::Boolean` or
    # `T.any(String, NilClass)` instead of `T.nilable(String)`.
    #
    # This is the inverse of `#simplify`.
    #
    # @abstract
    #: -> Type
    def normalize; end
   ```

2. Type simplification: transforms a type in an easy to read version.

   ```rb
    # Returns a simplified version of the type.
    #
    # Simplified types are meant to be easier to read, not to process.
    # For example, `T::Boolean` instead of `T.any(TrueClass, FalseClass)` or
    # `T.nilable(String)` instead of `T.any(String, NilClass)`.
    #
    # This is the inverse of `#normalize`.
    #
    # @abstract
    #: -> Type
    def simplify; end
   ```

We need type normalization to implement type simplification. Consider this example:

```rb
T.any(T.any(A, TrueClass, T.any(B, FalseClass, T.any(C, NilClass)))
```

The normalized version of this type is

```rb
T.any(A, TrueClass, B, FalseClass, C, NilClass)
```

Which can then be simplified into

```rb
T.nilable(A, B, C, T::Boolean)
```

### DIFF
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -2943,8 +2943,32 @@ class RBI::Type
   sig { returns(::RBI::Type) }
   def non_nilable; end
 
+  # Returns a normalized version of the type.
+  #
+  # Normalized types are meant to be easier to process, not to read.
+  # For example, `T.any(TrueClass, FalseClass)` instead of `T::Boolean` or
+  # `T.any(String, NilClass)` instead of `T.nilable(String)`.
+  #
+  # This is the inverse of `#simplify`.
+  #
+  # @abstract
+  sig { abstract.returns(::RBI::Type) }
+  def normalize; end
+
   sig { returns(::String) }
   def rbs_string; end
+
+  # Returns a simplified version of the type.
+  #
+  # Simplified types are meant to be easier to read, not to process.
+  # For example, `T::Boolean` instead of `T.any(TrueClass, FalseClass)` or
+  # `T.nilable(String)` instead of `T.any(String, NilClass)`.
+  #
+  # This is the inverse of `#normalize`.
+  #
+  # @abstract
+  sig { abstract.returns(::RBI::Type) }
+  def simplify; end
 
   # @abstract
   sig { abstract.returns(::String) }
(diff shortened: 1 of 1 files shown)


lib/rbi/type.rb
```diff
@@ -550,25 +845,8 @@ def nilable(type)
...
-
-        if flattened.size == 1
-          flattened.first #: as !nil
-        else
-          raise ArgumentError, "RBI::Type.all should have at least 2 types supplied" if flattened.size < 2
-
-          All.new(flattened)
-        end
+        type = All.new([type1, type2, *types])
+        type.simplify
```
paracycle, line 849:
```suggestion
        All.new([type1, type2, *types]).simplify
```
maybe?

lib/rbi/type.rb
```diff
@@ -577,64 +855,8 @@ def all(type1, type2, *types)
...
-          Any.new(types)
-        end
-
-        if is_nilable
-          nilable(type)
-        else
-          type
-        end
+        type = Any.new([type1, type2, *types])
+        type.simplify
```
paracycle, line 859:
```suggestion
        Any.new([type1, type2, *types]).simplify
```

lib/rbi/type.rb
```diff
@@ -673,9 +895,6 @@ def proc
         Proc.new
       end
 
-      # We mark the constructor as `protected` because we want to force the use of factories on `Type` to create types
-      protected :new
-
```
paracycle, line 678: Don't we still want this especially since `Type.any(Foo, Bar)` isn't necessarily the same as `Type::Any.new(Foo, Bar)` due to the `simplify` call in `any`, for instance?

PR #513, opened 2025-09-18
Resolve scopes of referenced types when merging trees

bcdanieldickison:
This is another attempt to fix tapioca's rbi incorrectly merging of the Stripe gem, as described here: https://github.com/Shopify/tapioca/issues/2248. Previous attempt: https://github.com/Shopify/rbi/pull/509

## Problem

When merging two trees that refer to the same scope using nested vs namespaced syntax, the merge algorithm was treating those as different scopes. For example:

```ruby
# lhs
module Foo; end
class Foo::Bar; end
class Foo::Baz < Foo::Bar; end

# rhs
module Foo
  class Bar; end; # See note 1
  class Baz < Bar; end
end
```

The two definitions of `Foo::Baz` were considered incompatible due to mismatched superclasses even though they technically both refer to `Foo::Bar`. *Note 1*: Additionally, in the case of the Stripe SDK, the superclass in right hand tree (the gem's exported rbi) could refer to a type only defined in the left hand tree (tapioca's generated rbi).

## This Solution

Wherever there's a type reference in the right hand tree during a merge, it is looked up in the left hand tree using rules mimicking Ruby's semantics. This is applied to the following places where types can be referenced:
* Class superclasses
* Mixins
* Signatures on Attrs and Methods

The output then uses fully-qualified names for those references.

Note that this makes the merge asymetric: the left-hand tree is expected to contain all the types that could be referenced by either tree.

## TODO
For further improvement, consider merging `Attr` and same-named `Method`. In the case of Stripe SDK, tapioca generates methods from source but the exported rbi defined attr accessors, which should be merged (along with their signatures). This is implemented in a second PR here: https://github.com/bandcampdotcom/rbi/pull/2

### DIFF
--- a/lib/rbi/rewriters/merge_trees.rb
+++ b/lib/rbi/rewriters/merge_trees.rb
@@ -228,20 +228,24 @@ module RBI
         def merge_nodes?(left, right)
           return false unless left.class == right.class
 
-          merge_comments(left, right) if left.is_a?(NodeWithComments)
+          merge_comments(left, right) if left.is_a?(NodeWithComments) && right.is_a?(NodeWithComments)
 
           case left
           when Class
+            right = right #: as Class
             left_superclass = lookup_type(name: left.superclass_name, referrer: left)
             right_superclass = lookup_type(name: right.superclass_name, referrer: right)
             left_superclass == right_superclass
           when Struct
+            right = right #: as Struct
             left.members == right.members && left.keyword_init == right.keyword_init
           when Const
+            right = right #: as Const
             left.name == right.name && left.value == right.value
           when Attr, Method
-            return false if left.is_a?(Method) && left.params != right.params
-            return false if left.is_a?(Attr) && left.names != right.names
+            right = right #: as Attr | Method
+            return false if left.is_a?(Method) && right.is_a?(Method) && left.params != right.params
+            return false if left.is_a?(Attr) && right.is_a?(Attr) && left.names != right.names
 
             left_sigs = left.sigs.map { fully_qualify_sig(_1, referrer: left) }
             right_sigs = right.sigs.map { fully_qualify_sig(_1, referrer: right) }
@@ -255,15 +259,19 @@ module RBI
               false
             end
           when Mixin
+            right = right #: as Mixin
             left_mixins = left.names.map { lookup_type(name: _1, referrer: left) }
             right_mixins = right.names.map { lookup_type(name: _1, referrer: right) }
             left_mixins == right_mixins
           when Helper
             # Do Helper names need to be resolved to types?
+            right = right #: as Helper
             left.name == right.name
           when Send
+            right = right #: as Send
             left.method == right.method && left.args == right.args
           when TStructField
+            right = right #: as TStructField
             left.name == right.name && left.type == right.type && left.default == right.default
           else
             true
@@ -274,17 +282,30 @@ module RBI
         # when referenced from the given referrer Node. The referrer can be
         # in a different tree, but its scope chain names will be used to find
         # the referent in this merge tree.
-        #: (name: String?, referrer: Node) -> Node?
+        #: (name: String?, referrer: Node) -> (Scope | Const)?
         def lookup_type(name:, referrer:)
           return unless name
 
-          return @index[name] if name.start_with?("::")
+          if name.start_with?("::")
+            referent = @index[name].last #: Node?
+            if referent.is_a?(Scope) || referent.is_a?(Const)
+              return referent
+            elsif referent
+              raise "Unexpected type #{referent} for #{name} with referrer #{referrer}"
+            else
+              return
+            end
+          end
 
-          referrer_scope = referrer.is_a?(Scope) ? referrer : referrer.parent_scope
+          referrer_scope = referrer.is_a?(Scope) ? referrer : referrer.parent_scope #: Scope?
           loop do
             scoped_name = "#{referrer_scope&.fully_qualified_name}::#{name}"
-            referent = @index[scoped_name].last
-            break referent if referent
+            referent = @index[scoped_name].last #: Node?
+            if referent.is_a?(Scope) || referent.is_a?(Const)
+              return referent
+            elsif referent
+              raise "Unexpected type #{referent} for #{name} with referrer #{referrer}"
+            end
             break unless referrer_scope
 
             referrer_scope = referrer_scope.parent_scope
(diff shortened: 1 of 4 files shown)


bcdanieldickison:
(Legal is reviewing the CLA currently.)

KaanOzkan (commented):
High level this solution makes sense to me. I can't immediately think of any issues but we'll test it. Let us know how CLA goes and also I triggered CI.

bcdanieldickison:
Thanks @KaanOzkan. I'm a bit stumped by the type checking error that remains, where I'm passing a Hash into a method using the keyword splat syntax:

```
lib/rbi/rewriters/merge_trees.rb:345: Expected `RBI::Type` but found `T::Hash[Symbol, RBI::Type]` for argument `params` https://srb.help/7002
     345 |            copy.params(**params) # This should be **params but sorbet seems to get tripped up?
                                  ^^^^^^^^
  Expected `RBI::Type` for argument `params` of method `RBI::Type::Proc#params`:
    lib/rbi/type.rb:736:
     736 |      #: (**Type params) -> self
                           ^^^^^^
```

Got any tips on how to express this typing correctly? It does function correctly despite the error.

lib/rbi/rewriters/merge_trees.rb
```diff
@@ -212,6 +213,169 @@ def replace_scope_header(left, right)
...
+            return false if left.is_a?(Method) && right.is_a?(Method) && left.params != right.params
+            return false if left.is_a?(Attr) && right.is_a?(Attr) && left.names != right.names
+
+            left_sigs = left.sigs.map { fully_qualify_sig(_1, referrer: left) }
+            right_sigs = right.sigs.map { fully_qualify_sig(_1, referrer: right) }
+            if left_sigs.empty? || right_sigs.empty? || left_sigs == right_sigs
+              right_sigs.each do |sig|
+                left_sigs << sig unless left_sigs.include?(sig)
+              end
+              left.sigs = left_sigs
```
KaanOzkan, line 256: Mutating in a `?` method doesn't feel right but maybe it's okay since it's simpler than alternatives.

lib/rbi/rewriters/merge_trees.rb
```diff
@@ -212,6 +213,169 @@ def replace_scope_header(left, right)
...
+        #: (name: String?, referrer: Node) -> (Scope | Const)?
+        def lookup_type(name:, referrer:)
+          return unless name
+
+          if name.start_with?("::")
+            referent = @index[name].last #: Node?
+            if referent.is_a?(Scope) || referent.is_a?(Const)
+              return referent
+            elsif referent
+              raise "Unexpected type #{referent} for #{name} with referrer #{referrer}"
```
KaanOzkan, line 294: I think this should raise a new `RBI::Merge::Error` similar to rest of the library raising custom errors.

lib/rbi/rewriters/merge_trees.rb
```diff
@@ -212,6 +213,169 @@ def replace_scope_header(left, right)
...
+          end
+
+          referrer_scope = referrer.is_a?(Scope) ? referrer : referrer.parent_scope #: Scope?
+          loop do
+            scoped_name = "#{referrer_scope&.fully_qualified_name}::#{name}"
+            referent = @index[scoped_name].last #: Node?
+            if referent.is_a?(Scope) || referent.is_a?(Const)
+              return referent
+            elsif referent
+              raise "Unexpected type #{referent} for #{name} with referrer #{referrer}"
```
KaanOzkan, line 307: Tested this on a repo that uses [spoom](https://github.com/Shopify/spoom/blob/main/lib/spoom/poset.rb#L22) and ran into `/Users/kaanozkan/src/github.com/Shopify/rbi/lib/rbi/rewriters/merge_trees.rb:307:in 'block in RBI::Rewriters::Merge::TreeMerger#lookup_type': Unexpected type ::Spoom::Poset::E for E with referrer ::Spoom::Poset#[](value) (RuntimeError)` for the generic type. For context, the RBS comments are translated into sigs at the start of tapioca execution https://github.com/Shopify/tapioca/blob/6bf27f3658e6c3a3b8757bb3079a8b04fbb7d10d/lib/tapioca/rbs/rewriter.rb#L50.

KaanOzkan (commented):
I tried to test the performance of this feature but ran into errors. I'd like to make sure it doesn't effect RBI generation too negatively.

lib/rbi/rewriters/merge_trees.rb
bcdanieldickison: Yeah I wasn't a fan of this either. Maybe the method can instead return `:compatible` (Symbol) rather than a boolean, and that way the linter won't demand the method be named with `?`.

Morriar (changes requested):
I don't think this is the right approach, we're shoving way too much logic into the merging and I'd prefer for it to stay as simple as possible: purely syntactic, no name resolution.

I think your previous approach was better but I would tweak it a bit:

1. Provide a flatten rewriter in `rbi` like you attempted in #509
2. Let Tapioca apply the flatten rewriter onto the imported RBI file
3. Merge the flattened RBIs together

That way we limit the complexity of names handling to the flatten rewriter and we do not need to touch the merge rewriter at all.

bcdanieldickison:
Thanks for the review @Morriar. Unfortunately, I don't think any flavor of purely syntactic merging is feasible. I wrote about it [in a comment](https://github.com/Shopify/rbi/pull/509#issuecomment-3300177969) when closing the other ticket, but the gist is that non-fully-qualified names can only be resolved by actually looking it up in the scope chain. E.g. when given this in the rhs:

```ruby
module Foo
  class Bar
    # could be ::Foo::Bar::Baz, ::Foo::Baz, or ::Baz
    sig { returns(Baz) }
    def baz; end
  end
end
```

it's ambiguous whether `Baz` refers to `::Foo::Bar::Baz`, `::Foo::Baz`, or `::Baz` without looking up types defined in the lhs. We also can't keep that sig return type as the unqualified `Baz` if we're merging this sig into a lhs tree that defined `Foo::Bar` without nesting, because that excludes `::Foo::Baz`:

```ruby
class Foo::Bar
  # can only be ::Foo::Bar::Baz or ::Baz
  sig { returns(Baz) }
  def baz; end
end
```

It we add the constraint that all type references must be defined in the same rbi tree, then I think a flatten rewriter can unambiguously resolve these types to fully-qualified names by only referencing each individual tree. Alas the Stripe gem rbi doesn't follow this convention—not sure if it's reasonable to say it's a bug with the exported rbi from the gem. I am curious how others are able to effectively use the Stripe gem with typings.

bcdanieldickison:
@KaanOzkan no problem. Yes, I think it's reasonable to ask Stripe to generate their exported rbi without nesting, using fully scoped names. We have an open issue that includes this problem https://github.com/stripe/stripe-ruby/issues/1577 so I'll forward this comment along there.

It'll probably be a good idea to explicitly disallow nested scoping in rbi files to avoid this problem elsewhere. I'm not sure if this should be at the rbi level or in tapioca, but would need to be a soft deprecation initially.

Morriar (commented):
<deleted>

Morriar:
I'm going to close this, to implement this properly we would need to keep track of the content of other RBI files. This seems more suited to something like https://github.com/Shopify/rubydex.

PR #585, opened 2026-04-09
Adopt `params` and `sig` from other when merging a `Method` with no sig

### LINKED ISSUE #2589: Tapioca v0.18.0 generates invalid gem RBI for stripe gem
Hi all,

I'm upgrading our codebase from Sorbet 0.5 and Tapioca 0.17.9 to Sorbet 0.6 and Tapioca 0.19.0.

After force regenerating all gem RBIs, Sorbet reports 24418 (!) errors in `sorbet/rbi/gems/stripe@19.0.0.rbi`.

At a glance, it appears that the signatures for all write accessors throughout the stripe gem are invalid:
```
sorbet/rbi/gems/stripe@19.0.0.rbi:4316: Malformed sig. Type not specified for parameter value https://srb.help/5003
    4316 |  def expand=(value); end
                        ^^^^^
    sorbet/rbi/gems/stripe@19.0.0.rbi:4315: Add it to the signature's params here
    4315 |  sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
```

The stripe gem provides its own RBI signatures, and the signatures for write accessors use `_name` as the argument name, e.g.:
```ruby
    sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
    def expand=(_expand); end
```

(from https://github.com/stripe/stripe-ruby/blob/v19.0.0/rbi/stripe/params/account_capability_list_params.rbi#L10-L11)

However, for some reason Tapioca forcefully replaces the argument name with `value` in the method definition, but not in the signature:
```ruby
  # pkg:gem/stripe#lib/stripe/params/account_capability_list_params.rb:7
  sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
  def expand=(value); end
```

At first I thought this might be related to #2248, but this appears to be a different issue. AFAICT, this is caused by a behavior change in Tapioca v0.18.0. When I force-regenerate the RBI with Tapioca v0.17.9, everything's fine.

I can work around it by not regenerating the RBI for the stripe gem for the time being, but this only works until we need to upgrade the stripe gem.

KaanOzkan:
Resolves https://github.com/Shopify/tapioca/issues/2589

When merging two method definitions where the left has no sig but the right does, adopt the right's param names alongside its sigs. Previously only the sigs were adopted, causing a mismatch between the method's param names and the sig's param names.

### DIFF
--- a/test/rbi/rewriters/merge_trees_test.rb
+++ b/test/rbi/rewriters/merge_trees_test.rb
@@ -1333,5 +1333,30 @@ module RBI
       RBI
       refute_empty(res.conflicts)
     end
+
+    def test_merge_methods_without_sig_adopts_params_and_sig_from_other
+      tree1 = parse_rbi(<<~RBI)
+        class Foo
+          def expand=(value); end
+        end
+      RBI
+
+      tree2 = parse_rbi(<<~RBI)
+        class Foo
+          sig { params(_a_different_name: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
+          def expand=(_a_different_name); end
+        end
+      RBI
+
+      res = tree1.merge(tree2)
+
+      assert_equal(<<~RBI, res.string)
+        class Foo
+          sig { params(_a_different_name: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
+          def expand=(_a_different_name); end
+        end
+      RBI
+      assert_empty(res.conflicts)
+    end
   end
 end
--- a/lib/rbi/rewriters/merge_trees.rb
+++ b/lib/rbi/rewriters/merge_trees.rb
@@ -479,10 +479,12 @@ module RBI
       super
 
       # If self is the all-anonymous side (since compatible_with? ensures
-      # at most one side is), adopt other's non-anonymous param names and sigs.
-      if params.all?(&:anonymous?)
+      # at most one side is), or if self has no sigs but other does, adopt
+      # other's non-anonymous param names and sigs.
+      if params.all?(&:anonymous?) || (sigs.empty? && !other.sigs.empty?)
         @params = other.params.dup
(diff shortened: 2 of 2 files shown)


test/rbi/rewriters/merge_trees_test.rb
```diff
@@ -1333,5 +1333,30 @@ def m2(_a); end
...
+          def expand=(_expand); end
+        end
+      RBI
+
+      res = tree1.merge(tree2)
+
+      assert_equal(<<~RBI, res.string)
+        class Foo
+          sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
+          def expand=(_expand); end
```
KaanOzkan: Before this change we'd have `def expand=(value); end` instead.

test/rbi/rewriters/merge_trees_test.rb
```diff
@@ -1333,5 +1333,30 @@ def m2(_a); end
...
+      tree1 = parse_rbi(<<~RBI)
+        class Foo
+          def expand=(value); end
+        end
+      RBI
+
+      tree2 = parse_rbi(<<~RBI)
+        class Foo
+          sig { params(_expand: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
+          def expand=(_expand); end
```
amomchilov: Can we make the difference more obvious?


```suggestion
          sig { params(a_different_name: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
          def expand=(a_different_name); end
```

amomchilov (approved):

paracycle (approved):
Thanks for the fix, and sorry I missed this in the first place.

PR #595, opened 2026-04-23
Bump `tapioca` to 0.19.1 & remove support for EOL Ruby 3.2

KaanOzkan:
## Bump `tapioca` to 0.19.1

Fixes a CI failure in `bin/export-rbi --check-sync`:

```
NoMethodError: undefined method `has_rest' for #<T::Private::Methods::Signature>
  tapioca-0.19.0/lib/tapioca/gem/listeners/sorbet_signatures.rb:26:in `compile_signature'
```

Caused by a `sorbet-runtime` API change that 0.19.1 accounts for.

Also regenerates gem RBIs via `bin/tapioca gem --all` to pick up updated `tapioca`, `sorbet`, `spoom`, and other transitive deps.

## Drop Ruby 3.2, add 4.0

Ruby 3.2 reached end-of-life on 2026-03-31. Bumps `required_ruby_version` in the gemspec from `>= 3.2` to `>= 3.3`.

This also unblocks the transitive `parallel` 2.x update pulled in alongside `tapioca` 0.19.1 — `parallel` 2.0 requires Ruby `>= 3.3`, which broke the old 3.2 matrix job.

Adds `4.0` to the CI matrix so we validate against the Ruby version already used locally (`.ruby-version` is `4.0.1`).


.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ "3.2", "3.3", "3.4" ]
+        ruby: [ "3.3", "3.4", "4.0" ]
```
KaanOzkan, line 11: Updated `main` protection rules to require `4.0` and removed `3.2`

vinistock (approved):

PR #601, opened 2026-05-12
Fix RBS Record translation to include optional fields

Morriar:
## Summary

* RBS Record types with optional fields (e.g. `{ a: Integer, ?b: String }`) were silently dropped during translation to RBI
* The translator used `type.fields` which only returns required fields — switched to `type.all_fields` which includes both required and optional fields
* Optional fields are now wrapped in `T.nilable` to represent that they may not be present
* Already-nilable optional fields (e.g. `?c: Float?`) are not double-wrapped thanks to `Type.nilable` simplification

### DIFF
--- a/lib/rbi/rbs/type_translator.rb
+++ b/lib/rbi/rbs/type_translator.rb
@@ -82,7 +82,10 @@ module RBI
             proc.bind(translate(type.self_type)) if type.self_type
             proc
           when ::RBS::Types::Record
-            Type.shape(type.fields.map { |name, type| [name, translate(type)] }.to_h)
+            Type.shape(type.all_fields.to_h do |name, (value, required)|
+              translated = translate(value)
+              [name, required ? translated : Type.nilable(translated)]
+            end)
           when ::RBS::Types::Tuple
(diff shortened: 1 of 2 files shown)


SomalianIvan (approved):

lib/rbi/rbs/type_translator.rb
```diff
@@ -82,7 +82,10 @@ def translate(type)
             proc.bind(translate(type.self_type)) if type.self_type
             proc
           when ::RBS::Types::Record
-            Type.shape(type.fields.map { |name, type| [name, translate(type)] }.to_h)
+            Type.shape(type.all_fields.map do |name, (value, required)|
+              translated = translate(value)
+              [name, required ? translated : Type.nilable(translated)]
+            end.to_h)
```
paracycle:
```suggestion
            Type.shape(type.all_fields.to_h do |name, (value, required)|
              translated = translate(value)
              [name, required ? translated : Type.nilable(translated)]
            end)
```

paracycle (approved):

PR #608, opened 2026-06-01
Refactor `RBS::TypeTranslator` to provide instance methods

### LINKED ISSUE #924: Add rewriter option for erasing generics
## Motivation

```ruby
#!/usr/bin/ruby

require "sorbet-runtime"

class Box
  extend T::Generic
  
  Value = type_member
  
  def initialize(value)
    @value = value
  end
end

extend T::Sig

sig { returns(Box[Integer]) }

```

</details>

## Proposed solution

For this example input:

```ruby
#: [E]
class Box
  #: -> void
  def initialize
    @elems = [] #: T::Array[E]
  end
end

#: -> Box[Integer]
```

```diff
 class Box
- extend T::Generic
- 
- E = type_member
  
  sig { void }
  def initialize
    @elems = [] #: T::Array[E]
  end
 end
 
-sig { returns(Box[Integer]) }
+sig { returns(Box) }
 def box_of_int = Box.new(123) #: Box[Integer]
```

dejmedus:
- Part of https://github.com/Shopify/spoom/issues/924

Previously, `RBS::TypeTranslator` methods lived on the class (ex. `TypeTranslator.translate(type)`)
In order to be able to support translation configuration, like the in-progress `erase_generic_types` option,
this PR converts `TypeTranslator`'s singleton methods to instance methods and updates it's callers

> If reviewing diff on Github, consider using Hide Whitespace setting

### DIFF
--- a/lib/rbi/rbs/type_translator.rb
+++ b/lib/rbi/rbs/type_translator.rb
@@ -4,196 +4,201 @@
 module RBI
   module RBS
     class TypeTranslator
+      #: type rbs_type = ::RBS::Types::Alias
+      #| | ::RBS::Types::Bases::Any
+      #| | ::RBS::Types::Bases::Bool
+      #| | ::RBS::Types::Bases::Bottom
+      #| | ::RBS::Types::Bases::Class
+      #| | ::RBS::Types::Bases::Instance
+      #| | ::RBS::Types::Bases::Nil
+      #| | ::RBS::Types::Bases::Self
+      #| | ::RBS::Types::Bases::Top
+      #| | ::RBS::Types::Bases::Void
+      #| | ::RBS::Types::ClassSingleton
+      #| | ::RBS::Types::ClassInstance
+      #| | ::RBS::Types::Function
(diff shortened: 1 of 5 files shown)


lib/rbi/rbs/type_translator.rb
```diff
@@ -4,196 +4,201 @@
...
-          when ::RBS::Types::Union
-            Type.any(*type.types.map { |t| translate(t) })
-          when ::RBS::Types::UntypedFunction
-            Type.proc.params(arg0: Type.untyped).returns(Type.untyped)
-          when ::RBS::Types::Variable
-            Type.type_parameter(type.name)
-          else
-            type #: absurd
-          end
+          new.translate(type)
```
dejmedus, line 34: Left this for backwards compatibility, but can be removed once we switch everything over

paracycle (approved):
Reviewing with whitespace diff turned off wasn't that painful.

amomchilov (approved):

PR #611, opened 2026-06-08
Add `erase_generic_types` option to RBS translators

### LINKED ISSUE #924: Add rewriter option for erasing generics

dejmedus:
- part of https://github.com/Shopify/spoom/issues/924

Previously, `RBS::TypeTranslator` and `RBS::MethodTypeTranslator` always translated generic types. However, as [`sorbet-runtime` is unable to enforce them](https://sorbet.org/docs/stdlib-generics#generics-and-runtime-checks), they are superfluous in runtime contexts

This PR adds a default `false` `erase_generic_types` kwarg to both translator's constructors. When set to `true` generic types are dropped (`Foo[Bar]` to `Foo`) and type params are treated as `T.anything`

### DIFF
--- a/lib/rbi/rbs/method_type_translator.rb
+++ b/lib/rbi/rbs/method_type_translator.rb
@@ -45,8 +45,10 @@ module RBI
 
       #: (::RBS::MethodType) -> void
       def visit(type)
-        type.type_params.each do |param|
-          result.type_params << param.name
+        unless @options.erase_generic_types
+          type.type_params.each do |param|
+            result.type_params << param.name
+          end
         end
 
         visit_function_type(type.type)
--- a/lib/rbi/rbs/type_translator.rb
+++ b/lib/rbi/rbs/type_translator.rb
@@ -66,7 +66,8 @@ module RBI
         when ::RBS::Types::Bases::Void
           Type.void
         when ::RBS::Types::ClassSingleton
-          type_parameter = type.args.first ? translate(type.args.first) : nil
+          type_arg = type.args.first
+          type_parameter = translate(type_arg) if type_arg && !@options.erase_generic_types
           Type.class_of(Type.simple(type.name.to_s), type_parameter)
         when ::RBS::Types::ClassInstance
           translate_class_instance(type)
@@ -106,7 +107,11 @@ module RBI
         when ::RBS::Types::UntypedFunction
           Type.proc.params(arg0: Type.untyped).returns(Type.untyped)
         when ::RBS::Types::Variable
-          Type.type_parameter(type.name)
+          if @options.erase_generic_types
+            Type.anything
+          else
+            Type.type_parameter(type.name)
+          end
         else
           type #: absurd
         end
@@ -128,10 +133,14 @@ module RBI
 
       #: (::RBS::Types::ClassInstance) -> Type
       def translate_class_instance(type)
-        return Type.simple(type.name.to_s) if type.args.empty?
+        type_name = type.name.to_s
+        return Type.simple(type_name) if type.args.empty?
 
-        type_name = translate_t_generic_type(type.name.to_s)
-        Type.generic(type_name, *type.args.map { |arg| translate(arg) })
+        if @options.erase_generic_types
+          Type.simple(erase_t_generic_type(type_name))
+        else
+          Type.generic(translate_t_generic_type(type_name), *type.args.map { |arg| translate(arg) })
+        end
       end
 
       #: (::RBS::Types::Function) -> Type
@@ -181,32 +190,35 @@ module RBI
         proc
       end
 
+      RUNTIME_GENERIC_TYPES = [
+        "Array",
+        "Class",
+        "Enumerable",
+        "Enumerator",
+        "Enumerator::Chain",
+        "Enumerator::Lazy",
+        "Hash",
+        "Module",
+        "Set",
+        "Range",
+      ].freeze #: Array[String]
+
+      GENERIC_TYPE_TO_SORBET_GENERIC_TYPE = RUNTIME_GENERIC_TYPES.flat_map do |type|
+        [[type, "::T::#{type}"], ["::#{type}", "::T::#{type}"]]
+      end.to_h.freeze #: Hash[String, String]
+
+      SORBET_GENERIC_TYPE_TO_GENERIC_TYPE = RUNTIME_GENERIC_TYPES.flat_map do |type|
+        [["T::#{type}", "::#{type}"], ["::T::#{type}", "::#{type}"]]
+      end.to_h.freeze #: Hash[String, String]
+
       #: (String type_name) -> String
       def translate_t_generic_type(type_name)
-        case type_name.delete_prefix("::")
-        when "Array"
-          "::T::Array"
-        when "Class"
-          "::T::Class"
-        when "Enumerable"
-          "::T::Enumerable"
-        when "Enumerator"
-          "::T::Enumerator"
-        when "Enumerator::Chain"
-          "::T::Enumerator::Chain"
-        when "Enumerator::Lazy"
-          "::T::Enumerator::Lazy"
-        when "Hash"
-          "::T::Hash"
-        when "Module"
-          "::T::Module"
-        when "Set"
-          "::T::Set"
-        when "Range"
-          "::T::Range"
-        else
-          type_name
-        end
+        GENERIC_TYPE_TO_SORBET_GENERIC_TYPE.fetch(type_name, type_name)
+      end
+
+      #: (String type_name) -> String
+      def erase_t_generic_type(type_name)
+        SORBET_GENERIC_TYPE_TO_GENERIC_TYPE.fetch(type_name, type_name)
       end
     end
   end
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -1443,6 +1443,9 @@ class RBI::RBS::TypeTranslator
 
   private
 
+  sig { params(type_name: ::String).returns(::String) }
+  def erase_t_generic_type(type_name); end
+
   sig { params(type: ::RBS::Types::ClassInstance).returns(::RBI::Type) }
   def translate_class_instance(type); end
 
@@ -1465,8 +1468,11 @@ class RBI::RBS::TypeTranslator
   end
 end
 
+RBI::RBS::TypeTranslator::GENERIC_TYPE_TO_SORBET_GENERIC_TYPE = T.let(T.unsafe(nil), Hash)
 RBI::RBS::TypeTranslator::Options = RBI::RBS::MethodTypeTranslator::Options
+RBI::RBS::TypeTranslator::RUNTIME_GENERIC_TYPES = T.let(T.unsafe(nil), Array)
 RBI::RBS::TypeTranslator::RbsType = T.type_alias { T.any(::RBS::Types::Alias, ::RBS::Types::Bases::Any, ::RBS::Types::Bases::Bool, ::RBS::Types::Bases::Bottom, ::RBS::Types::Bases::Class, ::RBS::Types::Bases::Instance, ::RBS::Types::Bases::Nil, ::RBS::Types::Bases::Self, ::RBS::Types::Bases::Top, ::RBS::Types::Bases::Void, ::RBS::Types::ClassInstance, ::RBS::Types::ClassSingleton, ::RBS::Types::Function, ::RBS::Types::Interface, ::RBS::Types::Intersection, ::RBS::Types::Literal, ::RBS::Types::Optional, ::RBS::Types::Proc, ::RBS::Types::Record, ::RBS::Types::Tuple, ::RBS::Types::Union, ::RBS::Types::UntypedFunction, ::RBS::Types::Variable) }
+RBI::RBS::TypeTranslator::SORBET_GENERIC_TYPE_TO_GENERIC_TYPE = T.let(T.unsafe(nil), Hash)
 
 class RBI::RBSComment < ::RBI::Comment
   sig { params(other: ::Object).returns(T::Boolean) }
--- a/test/rbi/rbs/method_type_translator_test.rb
+++ b/test/rbi/rbs/method_type_translator_test.rb
@@ -195,13 +195,47 @@ module RBI
         assert_equal(Type.class_of(Type.simple("Foo"), Type.simple("Bar")), sig.return_type)
       end
 
+      def test_erase_generic_types_replaces_method_type_parameters
+        sig = translate(
+          "[T, U] (Array[T], T?, T | Integer, [T, Integer], singleton(Foo)[T]) { (T) -> U } -> U?",
+          Method.new("foo", params: [
+            ReqParam.new("array"),
+            ReqParam.new("value"),
+            ReqParam.new("fallback"),
+            ReqParam.new("pair"),
+            ReqParam.new("klass"),
+            BlockParam.new("block"),
+          ]),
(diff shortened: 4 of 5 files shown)


lib/rbi/rbs/method_type_translator.rb
```diff
@@ -7,9 +7,9 @@ class MethodTypeTranslator
       class Error < RBI::Error; end
 
       class << self
-        #: (Method, ::RBS::MethodType) -> Sig
-        def translate(method, type)
-          translator = new(method)
```
amomchilov, line 12: This method is just a shim for backwards-compatibility, so there's no need to also add this option here.

Users who want the new API can just switch to `.new`.

lib/rbi/rbs/method_type_translator.rb
```diff
@@ -18,17 +18,20 @@ def translate(method, type)
...
       attr_reader :result
 
-      #: (Method) -> void
-      def initialize(method)
+      #: (Method, ?erase_generic_types: bool) -> void
+      def initialize(method, erase_generic_types: false)
         @method = method
         @result = Sig.new #: Sig
-        @type_translator = TypeTranslator.new #: TypeTranslator
+        @erase_generic_types = erase_generic_types
```
amomchilov: Style nit, keep the ivars initialized from the params first and in-order

```suggestion
        @erase_generic_types = erase_generic_types
        @result = Sig.new #: Sig
        @type_translator = TypeTranslator.new #: TypeTranslator
```

lib/rbi/rbs/type_translator.rb
```diff
@@ -99,7 +105,7 @@ def translate(type)
         when ::RBS::Types::UntypedFunction
           Type.proc.params(arg0: Type.untyped).returns(Type.untyped)
         when ::RBS::Types::Variable
-          Type.type_parameter(type.name)
+          @erase_generic_types ? Type.untyped : Type.type_parameter(type.name)
```
amomchilov: Rewrite into `T.anything` rather than `T.untyped`. It conveys a more specific specific intent.

```suggestion
          @erase_generic_types ? Type.anything : Type.type_parameter(type.name)
```

test/rbi/rbs/method_type_translator_test.rb
```diff
@@ -186,6 +186,27 @@ def test_translate_type_params
...
+            ReqParam.new("item"),
+          ]),
+          erase_generic_types: true,
+        )
+
+        assert_empty(sig.type_params)
+        assert_equal(
+          [
+            SigParam.new("array", Type.simple("Array")),
+            SigParam.new("item", Type.untyped),
```
amomchilov: Will need to update the tests to match, e.g.

```suggestion
            SigParam.new("item", Type.anything),
```

test/rbi/rbs/type_translator_test.rb
```diff
@@ -87,13 +87,46 @@ def test_translate_class_instance_generics_as_t_types
...
+
+      def test_translate_class_instance_without_args_unaffected_by_erase_generic_types
+        assert_equal(Type.simple("Foo"), translate("Foo", erase_generic_types: true))
+        assert_equal(Type.simple("::Foo::Bar"), translate("::Foo::Bar", erase_generic_types: true))
+      end
+
+      def test_translate_variable
+        assert_equal(Type.type_parameter(:T), translate_variable(:T))
+        assert_equal(Type.type_parameter(:E), translate_variable(:E))
+      end
```
amomchilov: Oh interesting, I guess this wasn't covered directly before, but indirectly via test/rbi/rbs/method_type_translator_test.rb

test/rbi/rbs/type_translator_test.rb
```diff
@@ -87,13 +87,46 @@ def test_translate_class_instance_generics_as_t_types
...
         assert_equal(Type.class_of(Type.simple("::Foo::Bar")), translate("singleton(::Foo::Bar)"))
         assert_equal(Type.class_of(Type.simple("Foo"), Type.simple("Bar")), translate("singleton(Foo)[Bar]"))
       end
 
+      def test_translate_class_singleton_with_erase_generic_types
+        assert_equal(Type.class_of(Type.simple("Foo")), translate("singleton(Foo)", erase_generic_types: true))
+        assert_equal(
+          Type.class_of(Type.simple("Foo")),
+          translate("singleton(Foo)[Bar]", erase_generic_types: true),
+        )
```
amomchilov: Notably, the expected value is the same in both cases. You can emphasize that by extracting the variable in common:


```suggestion
        expected = Type.class_of(Type.simple("Foo"))
        
        assert_equal(expected, translate("singleton(Foo)", erase_generic_types: true))
        
        assert_equal(
          expected,
          translate("singleton(Foo)[Bar]", erase_generic_types: true),
        )
```

lib/rbi/rbs/method_type_translator.rb
dejmedus: I like that!

lib/rbi/rbs/type_translator.rb
dejmedus: Sounds good:) I wondered if `anything` would act differently from `untyped` since Sorbet seems to [specifically](https://github.com/sorbet/sorbet/blob/ee31a06a56abb0e3ae40a16fe1eec1ac3ca8619d/gems/sorbet-runtime/lib/types/types/type_variable.rb) [use](https://github.com/sorbet/sorbet/blob/46fba38fa10f2f6539002d096f0f1e55a9546d0b/gems/sorbet-runtime/lib/types/types/base.rb#L68) `untyped`, but outside of [a snippet](https://gist.quick.shopify.io/a68b877d-f5f4-43c7-a01a-221174448d85) that couldn't come from statically correct RBS anyways it seems to be all good

test/rbi/rbs/type_translator_test.rb
dejmedus: I consolidated the tests a bit

lib/rbi/rbs/type_translator.rb
```diff
@@ -174,32 +184,36 @@ def translate_function(type)
...
-          type_name
-        end
+        runtime_type_name = type_name.delete_prefix("::")
+        return "::T::#{runtime_type_name}" if RUNTIME_GENERIC_TYPES.include?(runtime_type_name)
+
+        type_name
+      end
+
+      #: (String type_name) -> String
+      def erase_t_generic_type(type_name)
```
dejmedus, line 220: Similarly to how we add the prefix to translate generic types, we probably want to strip prefixes and use the actual runtime type when erasing. Otherwise something like [`#: (String component_name, T::Array[Symbol] includes) -> void`](https://github.com/shop/world/blob/66c591e8e3ab7d5449c9590009f250241dbdb4ff/areas/core/shopify/components/platform/essentials/app/models/shop.rb#L415) can error at runtime

amomchilov:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/Shopify/rbi/611?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#611** <a href="https://app.graphite.com/github/pr/Shopify/rbi/611?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rbi/611?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#616** <a href="https://app.graphite.com/github/pr/Shopify/rbi/616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

rbi/rbi.rbi
```diff
@@ -1386,8 +1386,8 @@ RBI::Public::DEFAULT = T.let(T.unsafe(nil), RBI::Public)
 module RBI::RBS; end
 
 class RBI::RBS::MethodTypeTranslator
-  sig { params(method: ::RBI::Method).void }
-  def initialize(method); end
+  sig { params(method: ::RBI::Method, erase_generic_types: T::Boolean).void }
```
paracycle: I don't think the erasure should apply here. Method type variables are properly checked at runtime, so should be kept.

lib/rbi/rbs/type_translator.rb
```diff
@@ -107,7 +108,7 @@ def translate(type)
         when ::RBS::Types::UntypedFunction
           Type.proc.params(arg0: Type.untyped).returns(Type.untyped)
         when ::RBS::Types::Variable
-          Type.type_parameter(type.name)
+          @options.erase_generic_types? ? Type.anything : Type.type_parameter(type.name)
```
paracycle: Inline with my other comment, I think these should always stay

lib/rbi/rbs/type_translator.rb
```diff
@@ -67,7 +67,8 @@ def translate(type)
         when ::RBS::Types::Bases::Void
           Type.void
         when ::RBS::Types::ClassSingleton
-          type_parameter = type.args.first ? translate(type.args.first) : nil
+          first_arg = type.args.first
+          type_parameter = first_arg && !@options.erase_generic_types? ? translate(first_arg) : nil
```
paracycle:
```suggestion
          type_arg = type.args.first
          type_parameter = translate(type_arg) if type_arg && !@options.erase_generic_types?
```
since in the `else` case, `type_parameter` will be initialized to `nil` anyway:
```ruby
irb> foo = 42 if false
#=> nil

irb> foo
#=> nil
```

lib/rbi/rbs/type_translator.rb
```diff
@@ -129,10 +130,14 @@ def translate_type_alias(type)
...
+        type_name = type.name.to_s
+        return Type.simple(type_name) if type.args.empty?
 
-        type_name = translate_t_generic_type(type.name.to_s)
-        Type.generic(type_name, *type.args.map { |arg| translate(arg) })
+        if @options.erase_generic_types?
+          return Type.simple(erase_t_generic_type(type_name))
+        end
+
+        Type.generic(translate_t_generic_type(type_name), *type.args.map { |arg| translate(arg) })
```
paracycle:
```suggestion
        if @options.erase_generic_types?
          Type.simple(erase_t_generic_type(type_name))
        else
          Type.generic(translate_t_generic_type(type_name), *type.args.map { |arg| translate(arg) })      
        end

```

lib/rbi/rbs/type_translator.rb
```diff
@@ -182,32 +187,36 @@ def translate_function(type)
...
+        "Class",
+        "Enumerable",
+        "Enumerator",
+        "Enumerator::Chain",
+        "Enumerator::Lazy",
+        "Hash",
+        "Module",
+        "Set",
+        "Range",
+      ].freeze #: Array[String]
```
paracycle, line 204: Can we create two lookups from this so that our operations below always become lookups?

```suggestion
      RUNTIME_GENERIC_TYPES = [
        "Array",
        "Class",
        "Enumerable",
        "Enumerator",
        "Enumerator::Chain",
        "Enumerator::Lazy",
        "Hash",
        "Module",
        "Set",
        "Range",
      ].freeze #: Array[String]
      
      GENERIC_TYPE_TO_SORBET_GENERIC_TYPE = RUNTIME_GENERIC_TYPES.flat_map do |type|
        [[type, "::T::#{type}"], ["::#{type}", "::T::#{type}"]]
      end.to_h
      
      SORBET_GENERIC_TYPE_TO_GENERIC_TYPE = RUNTIME_GENERIC_TYPES.flat_map do |type|
        [["T::#{type}", "::#{type}"], ["::T::#{type}", "::#{type}"]]
      end.to_h
```

lib/rbi/rbs/type_translator.rb
```diff
@@ -182,32 +187,36 @@ def translate_function(type)
...
-          "::T::Set"
-        when "Range"
-          "::T::Range"
-        else
-          type_name
-        end
+        runtime_type_name = type_name.delete_prefix("::")
+        return "::T::#{runtime_type_name}" if RUNTIME_GENERIC_TYPES.include?(runtime_type_name)
+
+        type_name
```
paracycle: This becomes:

```suggestion
      GENERIC_TYPE_TO_SORBET_GENERIC_TYPE.fetch(type_name, type_name)  
```

lib/rbi/rbs/type_translator.rb
```diff
@@ -182,32 +187,36 @@ def translate_function(type)
...
+      #: (String type_name) -> String
+      def erase_t_generic_type(type_name)
+        runtime_type_name = type_name
+          .delete_prefix("::T::")
+          .delete_prefix("T::")
+
+        return runtime_type_name if RUNTIME_GENERIC_TYPES.include?(runtime_type_name)
+
+        type_name
       end
```
paracycle, line 222: This becomes:

```suggestion
      def erase_t_generic_type(type_name)
        SORBET_GENERIC_TYPE_TO_GENERIC_TYPE.fetch(type_name, type_name)
      end
```

rbi/rbi.rbi
amomchilov: I don't think they are: https://github.com/sorbet/sorbet/blob/2ff17933b180a3bb4f924588e88caf1811522752/gems/sorbet-runtime/lib/types/types/type_parameter.rb#L34-L36


```ruby
#!/usr/bin/ruby

require "sorbet-runtime"

extend T::Sig

sig do
  s = type_parameters(:R)
    .params(i: T.type_parameter(:R))
    .returns(T.type_parameter(:R))
  
  p T.type_parameter(:R) # => #<T::Types::TypeParameter:0x00000001028f34e8 @name=:R>
  
  s
end
def identity(i) = "not i"

p identity(123) # => "not i"
```

lib/rbi/rbs/type_translator.rb
dejmedus: Thank you! I've made all the changes except for the method type vars being discussed above

lib/rbi/rbs/type_translator.rb
paracycle: Ok, I agree that we can replace method type variables with `T.anyting`, but nitpick: Can we make this a normal `if`/`else`. Usage of the ternary operator is really rare in Ruby:

```suggestion
          if @options.erase_generic_types?
            Type.anything
          else
            Type.type_parameter(type.name)
          end
```

paracycle (approved):
Thanks, I have a small stylistic comment, otherwise, this looks good to me!

Morriar (approved):

PR #612, opened 2026-06-10
Preserve sig parameter comments

KaanOzkan:
To be used by https://github.com/Shopify/spoom/pull/901

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -93,7 +93,8 @@ module RBI
         raise ParseError.new(message, location)
       end
 
-      visitor = TreeBuilder.new(source, comments: result.comments, file: file)
+      comments_by_line = result.comments.to_h { |c| [c.location.start_line, c] }
+      visitor = TreeBuilder.new(source, comments_by_line: comments_by_line, file: file)
       visitor.visit(result.value)
       visitor.tree
     rescue ParseError => e
@@ -112,13 +113,13 @@ module RBI
     end
 
     class Visitor < Prism::Visitor
-      #: (String source, file: String, ?comments_by_line: Hash[Integer, Prism::Comment]?) -> void
-      def initialize(source, file:, comments_by_line: nil)
+      #: (String source, file: String, ?comments_by_line: Hash[Integer, Prism::Comment]) -> void
+      def initialize(source, file:, comments_by_line: {})
         super()
 
         @source = source
         @file = file
-        @comments_by_line = comments_by_line || {} #: Hash[Integer, Prism::Comment]
+        @comments_by_line = comments_by_line
       end
 
       private
@@ -161,16 +162,15 @@ module RBI
         !!(node.is_a?(Prism::ConstantPathNode) && node_string(node) =~ /(::)?T::Sig::WithoutRuntime/)
       end
 
-      #: (Prism::Node node) -> Array[Comment]
-      def node_comments(node)
+      #: (Prism::Node node, ?min_line: Integer?) -> Array[Comment]
+      def node_comments(node, min_line: nil)
         comments = []
 
-        start_line = node.location.start_line
-        start_line -= 1 unless @comments_by_line.key?(start_line)
+        start_line, end_line = comment_lookup_range(node, min_line: min_line)
 
         rbs_continuation = [] #: Array[Prism::Comment]
 
-        start_line.downto(1) do |line|
+        start_line.downto(end_line) do |line|
           comment = @comments_by_line[line]
           break unless comment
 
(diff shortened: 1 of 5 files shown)


test/rbi/parser_test.rb
```diff
@@ -318,6 +318,33 @@ def test_parse_sig_comments
...
+        def foo(a, b); end
+      RBI
+
+      out = Parser.parse_string(rbi)
+      assert_equal(<<~RBI, out.string)
+        sig do
+          params(
+            a: Integer, # `a` comment
+            b: String # `b` comment 1
+                      # `b` comment 2
```
KaanOzkan, line 341: I opted to keep the existing RBI printer behaviour here so the comments are printed as trailing but we could also make it multiline like the RBS printer.

lib/rbi/parser.rb
```diff
@@ -112,12 +112,13 @@ def parse(source, file:)
     end
 
     class Visitor < Prism::Visitor
-      #: (String source, file: String) -> void
-      def initialize(source, file:)
+      #: (String source, file: String, ?comments_by_line: Hash[Integer, Prism::Comment]?) -> void
+      def initialize(source, file:, comments_by_line: nil)
```
Morriar: Can't we use a default here and avoid the `|| {}` lower?

```suggestion
      #: (String source, file: String, ?comments_by_line: Hash[Integer, Prism::Comment]) -> void
      def initialize(source, file:, comments_by_line: {})
```

lib/rbi/parser.rb
```diff
@@ -112,12 +112,13 @@ def parse(source, file:)
...
         super()
 
         @source = source
         @file = file
+        @comments_by_line = comments_by_line || {} #: Hash[Integer, Prism::Comment]
```
Morriar: And then:

```suggestion
        @comments_by_line = comments_by_line
```

lib/rbi/parser.rb
```diff
@@ -170,9 +241,12 @@ class TreeBuilder < Visitor
 
       #: (String source, comments: Array[Prism::Comment], file: String) -> void
       def initialize(source, comments:, file:)
-        super(source, file: file)
+        super(
+          source,
+          comments_by_line: comments.to_h { |c| [c.location.start_line, c] },
```
Morriar: Could we change https://github.com/Shopify/rbi/pull/612/changes#diff-161d1bda8afaf9a70d787f4ebe169196e46bd8c97114d3513afbf6135cb4c3f9R96 to directly pass `comments_by_line`? I think this way we could just call `super` directly.

lib/rbi/parser.rb
```diff
@@ -916,11 +921,16 @@ class SigBuilder < Visitor
...
-      #: (String content, file: String) -> void
-      def initialize(content, file:)
-        super
+      # Bounds sig param comment lookup to comments inside the current `params(...)` call.
+      #: Integer?
+      attr_reader :params_start_line
+
+      #: (String content, comments_by_line: Hash[Integer, Prism::Comment], file: String) -> void
+      def initialize(content, comments_by_line:, file:)
+        super(content, comments_by_line: comments_by_line, file: file)
```
Morriar: Shouldn't this work?

```suggestion
        super
```

lib/rbi/parser.rb
```diff
@@ -159,6 +160,76 @@ def self?(node)
...
         !!(node.is_a?(Prism::ConstantPathNode) && node_string(node) =~ /(::)?T::Sig::WithoutRuntime/)
       end
+
+      #: (Prism::Node node, ?min_line: Integer?) -> Array[Comment]
+      def node_comments(node, min_line: nil)
+        comments = []
+
+        start_line = node.location.start_line
+        start_line -= 1 unless @comments_by_line.key?(start_line)
+        start_line = [start_line, min_line].max if min_line
```
Morriar: I'm not sure I follow the new logic here.

Isn't the idea to lookup from the param's node line start down to where the params start to avoid going back to 1?

So shouldn't we do something like `start_line.downto(min_line) do |line|`?

KaanOzkan: Logic is complex due to edge cases, I refactored a bit, added more tests and comments to explain it better. Lmk what you think.

lib/rbi/parser.rb
```diff
@@ -168,11 +266,14 @@ class TreeBuilder < Visitor
...
       attr_reader :last_node
 
-      #: (String source, comments: Array[Prism::Comment], file: String) -> void
-      def initialize(source, comments:, file:)
-        super(source, file: file)
+      #: (String source, file: String, ?comments: Array[Prism::Comment]?, ?comments_by_line: Hash[Integer, Prism::Comment]) -> void
+      def initialize(source, file:, comments: nil, comments_by_line: {})
+        if comments
+          comments_by_line = comments.to_h { |comment| [comment.location.start_line, comment] }
+        end
```
KaanOzkan, line 276: `bundle exec spoom srb sigs export` [instantiates](https://github.com/Shopify/spoom/blob/7583bb1874d290b79c662a8c0eb998ae1eabcabd/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb#L153) `TreeBuilder` with `comments` so for now we need to keep compatibility.

lib/rbi/parser.rb
```diff
@@ -159,6 +161,104 @@ def self?(node)
       def t_sig_without_runtime?(node)
         !!(node.is_a?(Prism::ConstantPathNode) && node_string(node) =~ /(::)?T::Sig::WithoutRuntime/)
       end
+
+      #: (Prism::Node node, ?min_line: Integer?) -> Array[Comment]
+      def node_comments(node, min_line: nil)
```
KaanOzkan, line 166: This is mostly a move from `TreeBuilder` to `Visitor` so `SigBuilder` can use it too. It also gained a `min_line` parameter so callers can bound how far upward node comment lookup will scan.

lib/rbi/parser.rb
```diff
@@ -159,6 +161,104 @@ def self?(node)
...
+        comments = []
+
+        start_line, end_line = comment_lookup_range(node, min_line: min_line)
+
+        rbs_continuation = [] #: Array[Prism::Comment]
+
+        start_line.downto(end_line) do |line|
```
Morriar, line 173: Shouldn't this be `end_line.downto(start_line)`?

KaanOzkan: Sorry, I missed this notification.

We start from the node (`foo: Integer`) and scan the lines above it. So `start_line` is later in the file than `end_line`. I renamed them to make that clearer.

KaanOzkan: To simplify things further I got rid of the helper that calculated the lookup start/end lines and inlined a simpler version.

PR #616, opened 2026-06-16
Add `Options` object

### LINKED ISSUE #924: Add rewriter option for erasing generics

amomchilov:
- counterpart to https://github.com/Shopify/spoom/pull/954

This PR an Options object which will hold the shape of type translation config, including the soon to be added [`erase_generic_types` option](https://github.com/Shopify/spoom/issues/924)

### DIFF
--- a/lib/rbi/rbs/method_type_translator.rb
+++ b/lib/rbi/rbs/method_type_translator.rb
@@ -15,14 +15,32 @@ module RBI
         end
       end
 
+      class Options
+        #: bool
+        attr_reader :erase_generic_types
+
+        #: (?erase_generic_types: bool) -> void
+        def initialize(erase_generic_types: false)
+          @erase_generic_types = erase_generic_types
+        end
+
+        @default = new.freeze #: Options
+        class << self
+          #: Options
+          attr_reader :default
+        end
+      end
+
       #: Sig
       attr_reader :result
 
-      #: (Method) -> void
-      def initialize(method)
+      #: (Method, ?options: Options) -> void
+      def initialize(method, options: Options.default)
         @method = method
+        @options = options
+
         @result = Sig.new #: Sig
-        @type_translator = TypeTranslator.new #: TypeTranslator
+        @type_translator = TypeTranslator.new(options: @options) #: TypeTranslator
       end
 
       #: (::RBS::MethodType) -> void
--- a/lib/rbi/rbs/type_translator.rb
+++ b/lib/rbi/rbs/type_translator.rb
@@ -35,6 +35,13 @@ module RBI
         end
       end
 
+      Options = MethodTypeTranslator::Options
+
+      #: (?options: Options) -> void
+      def initialize(options: Options.default)
+        @options = options
+      end
+
       #: (rbs_type) -> Type
       def translate(type)
         case type
--- a/test/rbi/rbs/method_type_translator_test.rb
+++ b/test/rbi/rbs/method_type_translator_test.rb
@@ -200,7 +200,12 @@ module RBI
       #: (String, Method) -> RBI::Sig
       def translate(rbs_string, method)
         node = ::RBS::Parser.parse_method_type(rbs_string, require_eof: true)
-        RBS::MethodTypeTranslator.translate(method, node)
+
+        options = MethodTypeTranslator::Options.new
+
+        translator = RBS::MethodTypeTranslator.new(method, options:)
+        translator.visit(node)
+        translator.result
       end
     end
   end
--- a/test/rbi/rbs/type_translator_test.rb
+++ b/test/rbi/rbs/type_translator_test.rb
@@ -173,7 +173,8 @@ module RBI
       #: (String) -> RBI::Type
       def translate(rbs_string)
         node = ::RBS::Parser.parse_type(rbs_string, require_eof: true)
-        RBS::TypeTranslator.new.translate(node)
+        options = MethodTypeTranslator::Options.new
+        RBS::TypeTranslator.new(options:).translate(node)
       end
     end
   end
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -1386,8 +1386,8 @@ RBI::Public::DEFAULT = T.let(T.unsafe(nil), RBI::Public)
 module RBI::RBS; end
 
 class RBI::RBS::MethodTypeTranslator
-  sig { params(method: ::RBI::Method).void }
-  def initialize(method); end
+  sig { params(method: ::RBI::Method, options: ::RBI::RBS::MethodTypeTranslator::Options).void }
+  def initialize(method, options: T.unsafe(nil)); end
 
   sig { returns(::RBI::Sig) }
   def result; end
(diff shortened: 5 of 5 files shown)


amomchilov:
* **#611** <a href="https://app.graphite.com/github/pr/Shopify/rbi/611?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#616** <a href="https://app.graphite.com/github/pr/Shopify/rbi/616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/rbi/616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

amomchilov:
Needs `bundle exec spoom srb sigs export` but otherwise ready for review

KaanOzkan (approved):

lib/rbi/rbs/method_type_translator.rb
```diff
@@ -15,14 +15,35 @@ def translate(method, type)
...
 
+      # @abstract
+      class Options
+        #: (?erase_generic_types: bool) -> void
+        def initialize(erase_generic_types: false)
+          @erase_generic_types = erase_generic_types
+        end
+
+        #: -> bool
+        def erase_generic_types? = @erase_generic_types
```
Morriar: Where is `erase_generic_types?` used? I don't see any caller in the translate logic. Is the intent that a follow-up PR wires this up?

lib/rbi/rbs/method_type_translator.rb
```diff
@@ -15,14 +15,35 @@ def translate(method, type)
...
+        end
+      end
+
       #: Sig
       attr_reader :result
 
-      #: (Method) -> void
-      def initialize(method)
+      #: (Method, ?options: Options) -> void
+      def initialize(method, options: HumanReadableOptions.default)
```
Morriar: The class-level `MethodTypeTranslator.translate` and `TypeTranslator.translate` convenience methods don't forward options (`new(method)` / `new.translate(type)` use defaults). Should they accept an `options:` keyword too, or is the intent that callers needing custom options always instantiate directly?

lib/rbi/rbs/method_type_translator.rb
```diff
@@ -15,14 +15,35 @@ def translate(method, type)
...
+        #: (?erase_generic_types: bool) -> void
+        def initialize(erase_generic_types: false)
+          @erase_generic_types = erase_generic_types
+        end
+
+        #: -> bool
+        def erase_generic_types? = @erase_generic_types
+      end
+
+      class HumanReadableOptions < Options
```
Morriar: I think we're missing some documentation here explaining why we need this?

lib/rbi/rbs/method_type_translator.rb
dejmedus: I've updated the description with some extra context:) `erase_generic_types` is implemented in [this PR](https://github.com/Shopify/rbi/pull/611/changes#diff-042d21bb9fbbbea58018213d07562539e9a38a3d179d805852d1a0043e851cb5R111)

lib/rbi/rbs/method_type_translator.rb
dejmedus: Good point, I've added a comment explaining why HumanReadable is the default (mirroring [Spoom](https://github.com/Shopify/spoom/pull/954))

lib/rbi/rbs/method_type_translator.rb
dejmedus: > is the intent that callers needing custom options always instantiate directly

Yes, exactly! [decided here](https://github.com/Shopify/rbi/pull/611#discussion_r3377316784)

lib/rbi/rbs/method_type_translator.rb
```diff
@@ -15,14 +15,38 @@ def translate(method, type)
...
+      # Default options for formatted, human-readable translation output.
+      # Callers such as Spoom may also use these translators from line-matching
+      # rewrite modes
+      class HumanReadableOptions < Options
+        @default = new.freeze #: HumanReadableOptions
+        class << self
+          #: HumanReadableOptions
+          attr_reader :default
+        end
+      end
```
Morriar: This abstraction is leaking a concept that doesn't make sense in RBI, as your comment points out this is mostly for the client. Why dot move this one to `spoom`?

lib/rbi/rbs/type_translator.rb
```diff
@@ -35,6 +35,14 @@ def translate(type)
         end
       end
 
+      Options = MethodTypeTranslator::Options
+      HumanReadableOptions = MethodTypeTranslator::HumanReadableOptions
+
+      #: (?options: Options) -> void
+      def initialize(options: HumanReadableOptions.default)
```
Morriar: This should just be an `Options.default`. Let clients create and pass whatever subclass they want.

lib/rbi/rbs/method_type_translator.rb
dejmedus: Makes sense:) I've removed the concept of `HumanReadable` in favour of  just `Options` and an options default

test/rbi/rbs/method_type_translator_test.rb
```diff
@@ -200,7 +200,12 @@ def test_translate_generic_singleton
       #: (String, Method) -> RBI::Sig
       def translate(rbs_string, method)
         node = ::RBS::Parser.parse_method_type(rbs_string, require_eof: true)
-        RBS::MethodTypeTranslator.translate(method, node)
```
KaanOzkan, line 203: Why not have this helper accept options too so that it can be used here?

dejmedus: It was [decided](https://github.com/Shopify/rbi/pull/611#discussion_r3377316784) that callers who want to use options should have to instantiate directly

amomchilov: Yeah, it just becomes cumbersome to forward anything. The class method was a nice convenience when there were no options, but it makes less sense as we add more of them.

KaanOzkan (approved):

test/rbi/rbs/type_translator_test.rb
```diff
@@ -173,7 +173,8 @@ def test_translate_untyped_function
       #: (String) -> RBI::Type
       def translate(rbs_string)
         node = ::RBS::Parser.parse_type(rbs_string, require_eof: true)
-        RBS::TypeTranslator.new.translate(node)
+        options = MethodTypeTranslator::Options.new
+        RBS::TypeTranslator.new(options:).translate(node)
```
Morriar, line 177: I'm confused about the need for this change. Isn't the default value of `RBS::TypeTranslator.new` already an `Options.default` which is equivalent to `Options.new`?

test/rbi/rbs/method_type_translator_test.rb
```diff
@@ -200,7 +200,12 @@ def test_translate_generic_singleton
...
+
+        options = MethodTypeTranslator::Options.new
```
Morriar, line 204: I'm confused about the need for this change. Isn't the default value of `RBS::TypeTranslator.new` already an `Options.default` which is equivalent to `Options.new`?

PR #617, opened 2026-06-19
Backport `TypeTranslator` instance methods

amomchilov:
DO NOT MERGE. Just opened a PR for this branch for handy linking.

---

Adapted from https://github.com/Shopify/rbi/pull/608, which is on a much newer RBI version (0.8.14).

### DIFF
--- a/lib/rbi/version.rb
+++ b/lib/rbi/version.rb
@@ -2,5 +2,6 @@
 # frozen_string_literal: true
 
 module RBI
-  VERSION = "0.3.8"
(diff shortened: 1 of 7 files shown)


lib/rbi/version.rb
```diff
@@ -2,5 +2,6 @@
 # frozen_string_literal: true
 
 module RBI
-  VERSION = "0.3.14"
+  # This version is actually based on RBI v0.3.8, but we need to compare as `>= 0.3.15` for our backport use case.
```
amomchilov, line 5: This should actually read:


```suggestion
  # This version is actually based on RBI v0.3.8, but we need to compare as `>= 0.3.14` for our backport use case.
```

PR #624, opened 2026-07-08
Fix anonymous parameter modeling and merging

### LINKED ISSUE #928: RBS Rewriting produces invalid Ruby for anonymous block parameters
## Description

`Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs` generates invalid Ruby when translating an RBS block type annotation on a method that uses Ruby 3.1+ anonymous block forwarding (`&`). The generated `sig` contains `&block:` inside `params()`, which is a syntax error. Named block parameters (`&block`) are translated correctly.

## Reproduction

This code:

```ruby
require "spoom"

source = <<~RUBY
  # typed: true
  class Foo
    #: (String) ?{ (String) -> void } -> String
    def bar(request, &); end
  end
RUBY

rewritten = Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(source, file: "test.rb")
puts rewritten
RubyVM::InstructionSequence.compile(rewritten)
```

### Produces

```
# typed: true
class Foo
  sig { params(request: String, &block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
  def bar(request, &); end
end
(shopify):36:in 'RubyVM::InstructionSequence.compile': <compiled>:3: syntax errors found (SyntaxError)
  1 | # typed: true
  2 | class Foo
> 3 | ... : ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
    |     ^ unexpected ':', ignoring it
    |     ^ unexpected ':', expecting end-of-input
    |     ^ unexpected ':'; expected a `)` to close the arguments
    |                                                      ^ unexpected ')', ignoring it
    |                                                      ^ unexpected ')', expecting end-of-input
    |                                                       ^ unexpected '.', ignoring it
  4 |   def bar(request, &); end
  5 | end
```

### Expected

```
# typed: true
class Foo
  sig { params(request: String, block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
  def bar(request, &); end
end
=> <RubyVM::InstructionSequence:<compiled>@<compiled>:1>
```

## Environment

- Spoom 1.7.15
- Ruby 3.4.9
- Tapioca 0.19.1

Morriar:
## Summary

Fixes Shopify/spoom#928.

This updates RBI’s parameter model to represent Ruby anonymous parameters faithfully instead of approximating anonymity through parameter names.

Ruby supports anonymous rest, keyword rest, and block parameters:

```rb
def foo(*); end
def foo(**); end
def foo(&); end
```

Prism represents these with `name == nil`. RBI now preserves that information in the model instead of assigning synthetic names.

## Previous behavior

Before this change, every `RBI::Param` had a non-nil `name`, and anonymity was tracked through an `@anonymous` ivar derived from the name:

```rb
@anonymous = name.start_with?("_")
```

That meant names starting with `_` were treated as anonymous, even though they are still real Ruby parameter names.

However, this was name-based, not syntax-based. True Ruby anonymous parameters could not be represented directly because the model required a string name. The parser had to synthesize names instead:

```rb
def foo(*); end   # RestParam name became "*args"
def foo(**); end  # KwRestParam name became "**kwargs"
def foo(&); end   # BlockParam name became ""
```

Merge behavior was also somewhat surprising:

* `_`-prefixed names acted as wildcards for parameters of the same kind.
* Anonymous compatibility was all-or-nothing at the method level.
* Mixed anonymous/named parameter lists could conflict even when individual params were otherwise compatible.
* Merge could prefer/adopt parameter names from the other side depending on which side had signatures.
* The result could be asymmetric depending on which tree was merged into which.

## What changed

* `RBI::Param` is now more abstract and no longer stores a required `name`.
* Named parameter classes own non-nil names:
  * `ReqParam`
  * `OptParam`
  * `KwParam`
  * `KwOptParam`
* Optionally-anonymous parameter classes own nilable names:
  * `RestParam`
  * `KwRestParam`
  * `BlockParam`
* Anonymous params no longer carry an `@anonymous` ivar.
  * Anonymous-ness is derived from the parameter shape and whether the name is `nil`.
* Parser preserves `nil` names for anonymous Ruby params:
  * `*`
  * `**`
  * `&`
* Sig generation still uses printable projection names for anonymous params:
  * anonymous rest → `"*"`
  * anonymous kwrest → `"**"`
  * anonymous block → `"&"`
* Sig parsing/printing handles those quoted projection names correctly.

## Merge behavior

Method merging now treats anonymous params as compatible with named params only when the parameter kind is compatible.

Anonymous params are not a wildcard for arbitrary parameter shapes. They only mean “this parameter has no name.”

For example:

* `RestParam(nil)` is compatible with `RestParam("args")`
* `KwRestParam(nil)` is compatible with `KwRestParam("kwargs")`
* `BlockParam(nil)` is compatible with `BlockParam("block")`
* `RestParam(nil)` is not compatible with `ReqParam("arg")`

When one side has an anonymous parameter and the other side has a named parameter of the same kind, merge prefers the non-anonymous name.

Example:

```rb
sig { params("*": Integer, "**": T::Hash[Symbol, String], "&": T.proc.void).void }
def foo(*, **, &); end
```

merged with:

```rb
sig { params(args: Integer, kwargs: T::Hash[Symbol, String], block: T.proc.void).void }
def foo(*args, **kwargs, &block); end
```

produces:

```rb
sig { params(args: Integer, kwargs: T::Hash[Symbol, String], block: T.proc.void).void }
def foo(*args, **kwargs, &block); end
```

This also updates sig merging so anonymous sig projection names are renamed to match the final merged method parameters.

## Breaking changes

This is a breaking API change for consumers of the RBI model.

Previously, callers could generally assume every parameter had a non-nil name-like value. That is no longer true:

* `RBI::Param` itself no longer stores a required `name`.
* `RestParam#name`, `KwRestParam#name`, and `BlockParam#name` can now be `nil`.
* Anonymous parameters are no longer represented by synthesized names.
* `_`-prefixed parameter names are no longer considered anonymous by default.
* Code that previously compared anonymous params by placeholder names must now handle `nil` names or call `anonymous?`.

Merge behavior changed as well:

* anonymous-vs-named params of the same kind can now merge
* the merged method prefers the non-anonymous name
* anonymous params are not considered compatible across different parameter kinds
* `_`-prefixed named params no longer act as anonymous wildcards

This makes the model more accurate, but downstream code assuming anonymous params have stable synthetic names, or assuming `_`-prefixed names are anonymous, will need to be updated.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -599,14 +599,36 @@ module RBI
 
   # @abstract
   class Param < NodeWithComments
+    #: (?loc: Loc?, ?comments: Array[Comment]?) -> void
+    def initialize(loc: nil, comments: nil)
+      super(loc: loc, comments: comments)
+    end
+
+    #: -> String?
+    def name
+      case self
+      when ReqParam, OptParam, KwParam, KwOptParam, RestParam, KwRestParam, BlockParam
+        name
+      end
+    end
+
+    # @abstract
+    #: -> bool
+    def anonymous?
+      raise NotImplementedError, "Abstract method called"
+    end
+  end
+
+  class ReqParam < Param
+    # @override
     #: String
     attr_reader :name
 
-    #: (String name, ?loc: Loc?, ?comments: Array[Comment]?) -> void
-    def initialize(name, loc: nil, comments: nil)
+    #: (String name, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (ReqParam node) -> void } -> void
+    def initialize(name, loc: nil, comments: nil, &block)
       super(loc: loc, comments: comments)
       @name = name.to_s #: String
-      @anonymous = name.start_with?("_") #: bool
+      block&.call(self)
     end
 
     # @override
@@ -615,42 +637,61 @@ module RBI
       name
     end
 
+    # @override
     #: -> bool
     def anonymous?
-      @anonymous
+      false
     end
 
     #: (Object? other) -> bool
     def ==(other)
-      self.class === other &&
-        (name == other.name || anonymous? || other.anonymous?)
-    end
-  end
-
-  class ReqParam < Param
-    #: (String name, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (ReqParam node) -> void } -> void
-    def initialize(name, loc: nil, comments: nil, &block)
-      super(name, loc: loc, comments: comments)
-      block&.call(self)
+      ReqParam === other && (name == other.name)
     end
   end
 
   class OptParam < Param
+    # @override
+    #: String
+    attr_reader :name
+
     #: String
     attr_reader :value
 
     #: (String name, String value, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (OptParam node) -> void } -> void
     def initialize(name, value, loc: nil, comments: nil, &block)
-      super(name, loc: loc, comments: comments)
+      super(loc: loc, comments: comments)
+      @name = name.to_s #: String
       @value = value
       block&.call(self)
+    end
+
+    # @override
+    #: -> String
+    def to_s
+      "#{name} = #{value}"
+    end
+
+    # @override
+    #: -> bool
+    def anonymous?
+      false
+    end
+
+    #: (Object? other) -> bool
+    def ==(other)
+      OptParam === other && (name == other.name)
     end
   end
 
   class RestParam < Param
-    #: (String name, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (RestParam node) -> void } -> void
+    # @override
+    #: String?
+    attr_reader :name
+
+    #: (String? name, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (RestParam node) -> void } -> void
     def initialize(name, loc: nil, comments: nil, &block)
-      super(name, loc: loc, comments: comments)
+      super(loc: loc, comments: comments)
+      @name = name&.to_s #: String?
       block&.call(self)
     end
 
@@ -659,12 +700,28 @@ module RBI
     def to_s
       "*#{name}"
     end
+
+    # @override
+    #: -> bool
+    def anonymous?
+      name.nil?
+    end
+
+    #: (Object? other) -> bool
+    def ==(other)
+      RestParam === other && (name == other.name || anonymous? || other.anonymous?)
+    end
   end
 
   class KwParam < Param
+    # @override
+    #: String
+    attr_reader :name
+
     #: (String name, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (KwParam node) -> void } -> void
     def initialize(name, loc: nil, comments: nil, &block)
-      super(name, loc: loc, comments: comments)
+      super(loc: loc, comments: comments)
+      @name = name.to_s #: String
       block&.call(self)
     end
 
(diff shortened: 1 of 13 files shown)


lib/rbi/rbs_printer.rb
KaanOzkan: Agent found this failing test case after this change since `node.name` can be nil:

```rb
require "test_helper"

  class Foo
    def foo(**); end
  end
RBI

puts rbi.rbs_string
```
outputs 
```rbs
class Foo
  def foo: (**: untyped) -> untyped
end
```
which fails parsing, `RBS::ParsingError: a.rbs:2:14...2:15: Syntax error: unexpected token for simple type, token=`:` (pCOLON)`

lib/rbi/model.rb
```diff
@@ -673,51 +730,110 @@ def initialize(name, loc: nil, comments: nil, &block)
...
+      super(loc: loc, comments: comments)
+      @name = name&.to_s #: String?
       block&.call(self)
     end
 
     # @override
     #: -> String
     def to_s
-      "**#{name}:"
+      name ? "**#{name}" : "**"
```
rafaelfranca: This condition is unnecessary

lib/rbi/rbs_printer.rb
Morriar: Good catch!

lib/rbi/rewriters/merge_trees.rb
```diff
@@ -495,10 +501,47 @@ def merge_with(other)
...
+          end
+          SigParam.new(name, sig_param.type)
+        end
+        sig.params.replace(sig_params)
+      end
+    end
+  end
+
+  class Param
+    # @override
```
rafaelfranca, line 541: Override of what? This class has no subclass

rafaelfranca (approved):

lib/rbi/rewriters/merge_trees.rb
```diff
@@ -495,10 +501,47 @@ def merge_with(other)
...
+        sig_params = sig.params.zip(params).map do |sig_param, param|
+          param = param #: as !nil
+          name = if sig_param.anonymous? && !param.anonymous?
+            param.name #: as !nil
+          else
+            sig_param.name
+          end
+          SigParam.new(name, sig_param.type)
+        end
+        sig.params.replace(sig_params)
```
amomchilov, line 535: Can use `map!` instead

lib/rbi/model.rb
```diff
@@ -615,42 +637,61 @@ def to_s
...
+    attr_reader :name
+
     #: String
     attr_reader :value
 
     #: (String name, String value, ?loc: Loc?, ?comments: Array[Comment]?) ?{ (OptParam node) -> void } -> void
     def initialize(name, value, loc: nil, comments: nil, &block)
-      super(name, loc: loc, comments: comments)
+      super(loc: loc, comments: comments)
+      @name = name.to_s #: String
```
amomchilov, line 663: It's already a string, likewise below


```suggestion
      @name = name
```

lib/rbi/rewriters/merge_trees.rb
Morriar: It's an override of `Node#compatible_with?`. 

`Param` and the other nodes in this files are all subclasses of `Node`, though they are just reopenings so we didn't specify the superclass again.

rafaelfranca: > though they are just reopenings so we didn't specify the superclass again.

We should never do that. If you are reopening keep the inheritance correct.

lib/rbi/parser.rb
```diff
@@ -977,11 +977,21 @@ def visit_call_node(node)
...
         )
       end
 
+      #: (Prism::Node node) -> String
+      def sig_param_name(node)
+        case node
+        when Prism::SymbolNode
+          node.value.to_s
+        else
+          node_string!(node).delete_suffix(":")
```
amomchilov, line 991: BTW This breaks for unusual sigs, like `params("a": Integer)`

I have a wip to fix this across the repo

amomchilov (commented):
Forgot to submit my draft review.  🤦 

Small comments.

PR #625, opened 2026-07-09
Introduce ForwardingParam for argument forwarding

Morriar:
Depends on https://github.com/Shopify/rbi/pull/624.

Ruby argument forwarding (`...`) is a distinct parameter form; expanding it to `*args, **kwargs, &block` loses source fidelity and makes round-tripping impossible.

This PR:

* Adds `RBI::ForwardingParam` and wires it through parsing, visitors, RBI/RBS printing, and generated RBI signatures.
* Skips `AddSigTemplates` for methods containing `...`, since Sorbet does not support signatures on argument-forwarding methods.
* Treats forwarding params conservatively during merge: `foo(...)` is compatible with `foo(...)`, but incompatible with expanded `*args, **kwargs, &block` shapes.

Closes https://github.com/Shopify/rbi/pull/562.
Closes https://github.com/Shopify/rbi/pull/578.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -547,6 +547,11 @@ module RBI
       params << BlockParam.new(name)
     end
 
+    #: -> void
+    def add_forwarding_param
+      params << ForwardingParam.new
+    end
+
     #: (
     #|   ?params: Array[SigParam]?,
     #|   ?return_type: (String | Type),
@@ -607,7 +612,7 @@ module RBI
     #: -> String?
     def name
       case self
-      when ReqParam, OptParam, KwParam, KwOptParam, RestParam, KwRestParam, BlockParam
+      when ReqParam, OptParam, KwParam, KwOptParam, RestParam, KwRestParam, BlockParam, ForwardingParam
         name
       end
     end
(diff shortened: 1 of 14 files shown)


lib/rbi/model.rb
```diff
@@ -837,6 +842,37 @@ def ==(other)
     end
   end
 
+  class ForwardingParam < Param
```
amomchilov, line 846:
```suggestion
  # The `...` in `def m(...)`
  class ForwardingParam < Param
```

lib/rbi/rewriters/merge_trees.rb
```diff
@@ -467,6 +467,9 @@ def compatible_with?(other)
       return false unless other.is_a?(Method)
       return false unless name == other.name
       return false unless params.size == other.params.size
+      if (params_have_forwarding?(params) || params_have_forwarding?(other.params)) && params != other.params
```
amomchilov, line 470: Why `params != other.params`?

Morriar: Consider these methods:

 ```ruby
   def foo(a, ...); end
 ```

 vs

 ```ruby
   sig { params(a: Integer, rest: T.untyped).void }
   def foo(a, *rest); end
 ```

Should we merge them?

amomchilov: Ah yes, makes sense.

In a future where Sorbet actually supported sigs for `...`, then we could revise to:

```ruby
  sig { params(a: Integer, rest: T.untyped, "...": T.untyped).void }
  def foo(a, *rest, ...); end
```

amomchilov (approved):

PR #633, opened 2026-07-21
Preserve string-ness of string keys in Shape#to_rbi

tavianator:
`RBI::Type::Shape#to_rbi` rendered shapes like `key: Value`
unconditionally, which turned string keys into symbol keys.  Fix it by
checking for string keys and rendering those like `"key" => Value`.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/test/rbi/type_test.rb
+++ b/test/rbi/type_test.rb
@@ -259,7 +259,7 @@ module RBI
         :qux => Type.simple("String"),
       )
       refute_predicate(type, :nilable?)
-      assert_equal("{ foo: String, bar: String, baz: String, qux: String }", type.to_rbi)
+      assert_equal("{ foo: String, bar: String, \"baz\" => String, qux: String }", type.to_rbi)
 
       type = Type.shape(
         "\"foo\"": Type.simple("String"),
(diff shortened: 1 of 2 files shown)


juharris (approved):

KaanOzkan (approved):
Thanks.

test/rbi/type_test.rb
```diff
@@ -259,15 +259,24 @@ def test_build_type_shape
...
+      assert_equal("{ foo: String, bar: String, \"baz\" => String, qux: String }", type.to_rbi)
 
       type = Type.shape(
         "\"foo\"": Type.simple("String"),
         "\"bar\"" => Type.simple("String"),
         :"\"baz\"" => Type.simple("String"),
       )
       refute_predicate(type, :nilable?)
-      assert_equal("{ \"foo\": String, \"bar\": String, \"baz\": String }", type.to_rbi)
+      assert_equal("{ \"foo\": String, \"\\\"bar\\\"\" => String, \"baz\": String }", type.to_rbi)
```
amomchilov, line 270: Using `%q{}` here would let us not need all the `\`

PR #643, opened 2026-08-04
Support multiple type parameters on T.class_of

### LINKED ISSUE #642: `T.class_of(Foo)[...]` with a `type_template` fails to parse (expects exactly 1 type argument)
## Summary

`T.class_of(Foo)[...]` with more than one type argument cannot be parsed, so any singleton class with a user-defined `type_template` fails to translate to RBS.

Sorbet's rule ([docs](https://sorbet.org/docs/class-of#tclass_of-applying-type-arguments-to-a-singleton-class-type)): the first bracket argument is the attached class, and the remaining arguments apply to each `type_template` the class defines, in turn, e.g. `T.class_of(AnotherClass)[AnotherClass, Integer]`. `rbi` hardcodes exactly one argument.

## Repro

```ruby
# typed: true
class Step
  class << self
    extend T::Generic
    ContextType = type_template
  end
end

class Base
  extend T::Sig

  sig { params(step: T.class_of(Step)[Step, Integer]).void }
  def self.run(step); end
end
```

`spoom srb sigs translate` (spoom 1.8.2):

```
Warning: Can't parse ./base.rb: Failed to parse type T.class_of(Step)[Step, Integer] (Expected exactly 1 argument, got 2)
```

Reduced to `rbi` alone (main, 5462b69, v0.4.2):

```ruby
require "rbi"

RBI::Type.parse_string("T.class_of(Step)[Step]")          # => singleton(Step)[Step]
RBI::Type.parse_string("T.class_of(Step)[Step, Integer]") # RBI::Type::Error: Expected exactly 1 argument, got 2

RBI::Parser.parse_string(
  %{sig { params(step: T.class_of(Step)[Step, Integer]).void }\ndef foo(step); end\n}
).rbs_string
# RBI::RBSPrinter::Error: Failed to parse type `T.class_of(Step)[Step, Integer]` (Expected exactly 1 argument, got 2)
```

## Cause

`lib/rbi/type_parser.rb`, in the `T.class_of(Foo)[Bar]` branch:

```ruby
type_param_args = check_arguments_exactly!(node, 1)
type_param = parse_node(type_param_args.first)
return Type::ClassOf.new(type, type_param)
```

The arity is fixed at 1, and `Type::ClassOf` only holds a single type param, so there is nowhere to put the `type_template` arguments even if the parser accepted them. Note the RBS side needs a decision too: `singleton(Step)` is not generic in RBS, so a faithful translation may not exist and the printer may need to drop the extra arguments (or the case may need to be reported rather than translated).

Morriar:
Fixes #642.

- allow `T.class_of(Foo)[...]` to parse more than one type parameter
- preserve all singleton type parameters in RBI and RBS output
- translate multiple RBS singleton type parameters back to RBI

### DIFF
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -2671,7 +2671,12 @@ class RBI::Type
     sig { returns(::RBI::Type::Boolean) }
     def boolean; end
 
-    sig { params(type: ::RBI::Type::Simple, type_parameters: ::RBI::Type).returns(::RBI::Type::ClassOf) }
+    sig do
+      params(
+        type: ::RBI::Type::Simple,
+        type_parameters: T.any(::RBI::Type, T::Array[::RBI::Type])
+      ).returns(::RBI::Type::ClassOf)
+    end
     def class_of(type, *type_parameters); end
 
     sig { params(name: ::String, params: T.any(::RBI::Type, T::Array[::RBI::Type])).returns(::RBI::Type::Generic) }
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -2671,8 +2671,8 @@ class RBI::Type
     sig { returns(::RBI::Type::Boolean) }
     def boolean; end
 
-    sig { params(type: ::RBI::Type::Simple, type_parameter: T.nilable(::RBI::Type)).returns(::RBI::Type::ClassOf) }
-    def class_of(type, type_parameter = T.unsafe(nil)); end
+    sig { params(type: ::RBI::Type::Simple, type_parameters: ::RBI::Type).returns(::RBI::Type::ClassOf) }
+    def class_of(type, *type_parameters); end
 
     sig { params(name: ::String, params: T.any(::RBI::Type, T::Array[::RBI::Type])).returns(::RBI::Type::Generic) }
     def generic(name, *params); end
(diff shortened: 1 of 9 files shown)


rbi/rbi.rbi
```diff
@@ -2671,8 +2671,8 @@ class RBI::Type
     sig { returns(::RBI::Type::Boolean) }
     def boolean; end
 
-    sig { params(type: ::RBI::Type::Simple, type_parameter: T.nilable(::RBI::Type)).returns(::RBI::Type::ClassOf) }
-    def class_of(type, type_parameter = T.unsafe(nil)); end
+    sig { params(type: ::RBI::Type::Simple, type_parameters: ::RBI::Type).returns(::RBI::Type::ClassOf) }
```
st0012: Does RBI support typing splat args? Do we need to say it's `RBI::Type` or `T::Array[RBI::Type]`?

st0012: Basically, should this part match `generic`'s sig?

Morriar: > Does RBI support typing splat args? Do we need to say it's `RBI::Type` or `T::Array[RBI::Type]`?

We don't repeat it in Sorbet, see [this example](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28args%3A%20Integer%29.void%20%7D%0Adef%20foo%28*args%29%0A%20%20T.reveal_type%28args%29%0Aend).

> Basically, should this part match generic's sig?

But you're right we should match the API of `Generic`. Updated 👍

st0012 (approved):

PR #644, opened 2026-08-04
Fix RBS translation of braceless shape return types

### LINKED ISSUE #641: RBS translation fails for brace-less shape types in `sig { returns(...) }`
## Summary

A `sig` whose `returns` takes a **brace-less hash** (a shape type written without `{}`) fails to translate to RBS, because the sig's type is stored as a source slice and re-parsed out of context. Standalone, `"a" => Integer` is rightward pattern matching (`Prism::MatchRequiredNode`), not a hash.

Both `RBI::Type` and `RBI::RBSPrinter` support shapes with string keys, so this is purely a round-trip problem: adding braces makes the exact same type translate fine.

## Repro

```ruby
# typed: true
class Foo
  extend T::Sig

  sig { returns("a" => Integer) }
  def foo = { "a" => 1 }
end
```

`spoom srb sigs translate` (spoom 1.8.2):

```
Warning: Can't parse ./foo.rb: Failed to parse type "a" => Integer (Unexpected node #<Prism::MatchRequiredNode:0x000000012e191908>)
```

Reduced to `rbi` alone (main, 5462b69, v0.4.2):

```ruby
require "rbi"

RBI::Parser.parse_string(%{sig { returns("a" => Integer) }\ndef foo; end\n}).rbs_string
# RBI::RBSPrinter::Error: Failed to parse type `"a" => Integer` (Unexpected node `#<Prism::MatchRequiredNode>`)

RBI::Parser.parse_string(%{sig { returns(a: Integer) }\ndef foo; end\n}).rbs_string
# RBI::RBSPrinter::Error: Failed to parse type `a: Integer` (unexpected ':', expecting end-of-input. unexpected ':', ignoring it.)

RBI::Parser.parse_string(%{sig { returns({"a" => Integer}) }\ndef foo; end\n}).rbs_string
# => "def foo: -> {\"a\" => Integer}"   # works
```

The symbol-key form `returns(a: Integer)` is affected too, it just fails with a syntax error instead of a `MatchRequiredNode`.

`RBI::Type.parse_string` shows the same split:

```ruby
RBI::Type.parse_string(%q{"a" => Integer})   # raises
RBI::Type.parse_string(%q[{"a" => Integer}]) # => {"a" => Integer}
```

## Cause

- `lib/rbi/parser.rb` (`when "returns"`) stores the return type as `node_string!(first)`. When the argument is a brace-less hash, `first` is a `Prism::KeywordHashNode` and its slice has no braces.
- `lib/rbi/rbs_printer.rb#parse_type` later calls `RBI::Type.parse_string` on that slice. Out of the call's argument context, Prism parses `"a" => Integer` as a `MatchRequiredNode`, which `RBI::Type.parse_node` rejects.

`params` is not affected: each value node carries its own braces.

Possible fixes: keep the `Prism` node (or a normalized string with braces) when the `returns` argument is a `KeywordHashNode`, or let `RBI::Type.parse_node` accept `MatchRequiredNode` / re-wrap the slice in braces before parsing.

## Found in

Shopify Core, `components/online_store/app/models/theme/settings_data/platform_customizations.rb`:

```ruby
sig do
  returns("custom_css" => Theme::SettingsData::CustomCss)
end
def segments
```

Morriar:
Fixes #641

- preserve braceless hash-rocket return shapes by restoring braces around Prism keyword hash nodes
- distinguish hash rockets from keyword arguments using association operator locations
- cover string, symbol, mixed-key, single-key, and multi-key shapes
- keep invalid keyword-argument forms rejected during RBS translation

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -956,7 +956,11 @@ module RBI
           args = node.arguments
           if args.is_a?(Prism::ArgumentsNode)
             first = args.arguments.first
-            @current.return_type = node_string!(first) if first
+            if first
+              return_type = node_string!(first)
+              return_type = "{#{return_type}}" if braceless_shape?(first)
+              @current.return_type = return_type
+            end
           end
         when "type_parameters"
           args = node.arguments
@@ -980,6 +984,15 @@ module RBI
           sig_param_name(node.key),
           node_string!(node.value),
         )
+      end
+
+      #: (Prism::Node node) -> bool
+      def braceless_shape?(node)
+        return false unless node.is_a?(Prism::KeywordHashNode)
+
+        node.elements.all? do |element|
+          element.is_a?(Prism::AssocNode) && element.operator_loc
+        end
       end
 
       #: (Prism::Node node) -> String
--- a/rbi/rbi.rbi
+++ b/rbi/rbi.rbi
@@ -1033,6 +1033,9 @@ class RBI::Parser::SigBuilder < ::RBI::Parser::Visitor
   sig { params(node: ::Prism::CallNode, value: ::String).returns(T::Boolean) }
   def allow_incompatible_override?(node, value); end
 
+  sig { params(node: ::Prism::Node).returns(T::Boolean) }
+  def braceless_shape?(node); end
+
   sig { returns(::RBI::Sig) }
   def current; end
 
--- a/test/rbi/parser_test.rb
+++ b/test/rbi/parser_test.rb
@@ -237,6 +237,50 @@ module RBI
       RBI
     end
 
+    def test_parse_sigs_with_braceless_shape_return_type
+      rbi = <<~RBI
+        sig { returns("foo" => Integer) }
+        def single; end
+
+        sig { returns(:foo => Integer) }
+        def symbol; end
+
+        sig { returns("foo" => Integer, :bar => String) }
+        def mixed_keys; end
+
+        sig { returns("foo" => Integer, "bar" => String) }
+        def multiple; end
+      RBI
+
+      tree = parse_rbi(rbi)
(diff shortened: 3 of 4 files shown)


lib/rbi/parser.rb
```diff
@@ -982,6 +986,15 @@ def visit_assoc_node(node)
         )
       end
 
+      #: (Prism::Node node) -> bool
+      def braceless_shape?(node)
+        return false unless node.is_a?(Prism::KeywordHashNode)
+
+        node.elements.all? do |element|
+          element.is_a?(Prism::AssocNode) && !element.operator_loc.nil?
```
paracycle: Nitpick:
```suggestion
          element.is_a?(Prism::AssocNode) && element.operator_loc
```
unless `operator_loc` can somehow be `false` and that would be a desired value for us.

lib/rbi/parser.rb
```diff
@@ -956,7 +956,11 @@ def visit_call_node(node)
           args = node.arguments
           if args.is_a?(Prism::ArgumentsNode)
             first = args.arguments.first
-            @current.return_type = node_string!(first) if first
+            if first
```
paracycle, line 959: Could the same situation not happen in a different `sig` method? For example, could it happen in `bind`?

paracycle (approved):

lib/rbi/parser.rb
Morriar: Good call:

1. https://github.com/Shopify/rbi/pull/645 (I actually had no idea this was valid, the `bind` inside `T.proc` is already handled)
2. https://github.com/Shopify/rbi/pull/647

Morriar (commented):
No comments from me. The parser change is small and covered by parser/RBS printer tests.

PR #645, opened 2026-08-04
Add support for top-level `bind` in sigs

Morriar:
## Summary

Add support for Sorbet’s top-level `bind` signature modifier:

```ruby
sig { bind(Foo).params(value: String).returns(Symbol) }
```

Previously, the parser silently discarded `bind(Foo)`.

- store the bind type on `RBI::Sig`
- parse and print `bind(...)` when round-tripping RBI
- normalize braceless shape bind types:

  ```ruby
  sig { bind("foo" => Integer).void }
  ```

  becomes:

  ```ruby
  sig { bind({"foo" => Integer}).void }
  ```

- include bind types in signature equality and merge behavior
- preserve bind types when rewriting attributes into methods
- share type-argument extraction between `bind` and `returns`

RBS output intentionally omits top-level bind. Sorbet uses it to set the type of `self` while checking the method body, and RBS has no equivalent method-level self binding.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -555,6 +555,7 @@ module RBI
     #: (
     #|   ?params: Array[SigParam]?,
     #|   ?return_type: (String | Type),
+    #|   ?bind_type: (String | Type)?,
     #|   ?is_abstract: bool,
     #|   ?is_override: bool,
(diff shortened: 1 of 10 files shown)


paracycle (approved):
I actually didn't know that you could do a top-level `bind` call like this either.

PR #646, opened 2026-08-04
Require `RBI::Type` objects throughout the model

Morriar:
## What

This PR removes the backwards-compatible support for passing type strings to RBI model objects. APIs that previously accepted `RBI::Type | String` now accept only `RBI::Type`.

It also migrates internal producers and consumers to the new contract:

- The parser converts type syntax into `RBI::Type` objects using `Type.parse_string`.
- Signature return types, signature parameters, and `T::Struct` fields are represented as `Type`.
- Rewriters and benchmarks construct types through the `Type` API.
- The RBI printer consumes `Type` objects directly.
- The RBS printer handles and simplifies already-parsed `Type` objects.

## Why

`RBI::Type` was introduced while retaining String support for backwards compatibility. Keeping both representations means every consumer must handle two possible types and allows unparsed type syntax to flow through the model.

Using a single representation:

- Makes the model contract explicit.
- Centralizes parsing and validation in the type parser.
- Removes String compatibility branches from downstream consumers.
- Allows rewriters and printers to operate on the structured type representation.

## Output changes

Parsing and printing an RBI now uses `Type#to_rbi`, which canonicalizes Sorbet helpers with an absolute `::T` prefix.

For example:

```diff
-sig { params(value: T.untyped).returns(T.nilable(String)) }
+sig { params(value: ::T.untyped).returns(::T.nilable(String)) }
```

This preserves the existing `Type#to_rbi` behavior and prevents a locally defined `T` constant from shadowing Sorbet’s top-level `T`.

Generic names whose qualification is represented directly by the parsed type remain unchanged, such as `T::Array[String]`.

## Breaking changes

Callers must replace type strings with `RBI::Type` objects:

```diff
-RBI::SigParam.new("value", "T.nilable(String)")
+RBI::SigParam.new(
+  "value",
+  RBI::Type.nilable(RBI::Type.simple("String")),
+)
```

```diff
-RBI::Sig.new(return_type: "void")
+RBI::Sig.new(return_type: RBI::Type.void)
```

The affected APIs include:

- `RBI::Method#add_sig`
- `RBI::Sig#initialize`
- `RBI::Sig#return_type=`
- `RBI::Sig#add_param`
- `RBI::SigParam#initialize`
- `RBI::TStructConst#initialize`
- `RBI::TStructProp#initialize`
- `RBI::TStructField#type=`


rbi/rbi.rbi
```diff
@@ -2324,7 +2327,7 @@ class RBI::Sig < ::RBI::NodeWithComments
   sig { returns(T::Array[::RBI::SigParam]) }
   def params; end
 
-  sig { returns(T.any(::RBI::Type, ::String)) }
+  sig { returns(::RBI::Type) }
   def return_type; end
 
   def return_type=(_arg0); end
```
KaanOzkan, line 2346: FYI I think this is coming from an `attr_accessor` above and even though the reader is typed the writer isn't.

KaanOzkan (approved):

PR #653, opened 2026-08-19
Preserve comments inside Sorbet signatures

Morriar:
## Summary

Preserve comments attached to parameters inside Sorbet signatures when parsing RBI files and printing RBS.

This work is based on and supersedes #612 by @KaanOzkan.

- Move shared comment parsing into the base parser visitor so `TreeBuilder` and `SigBuilder` consume the same comment map.
- Attach comments inside a signature to the closest representable node:
  - signature-level comments to `Sig#comments`
  - parameter comments to `SigParam#comments`
  - comments outside the signature source range to the method
- Bound parameter comment lookup to its containing `params(...)` call.
- Preserve inline and leading parameter comments without capturing comments from surrounding calls.
- Print ordinary RBS parameter comments at the end of their parameter line.
- Print block parameter comments at the end of the complete signature line.
- Keep block signatures inline after overload separators.
- Retain the legacy `TreeBuilder(..., comments:)` keyword used by Spoom.
- Regenerate the exported RBI.

## Example

Given:

```ruby
sig do
  params(
    # User identifier
    id: Integer,
    # Callback
    block: T.proc.void
  ).void
end
def foo(id, &block); end
```

RBS output becomes:

```rbs
def foo: (
  Integer id # User identifier
) { -> void } -> void # Callback
```

Overloads remain inline:

```rbs
def foo: -> void
       | { -> void } -> void # Callback
```

Multiple comments on an ordinary parameter remain aligned:

```rbs
Integer id, # User identifier
            # Generated externally
```

### DIFF
--- a/lib/rbi/parser.rb
+++ b/lib/rbi/parser.rb
@@ -93,7 +93,8 @@ module RBI
         raise ParseError.new(message, location)
       end
 
-      visitor = TreeBuilder.new(source, comments: result.comments, file: file)
+      comments_by_line = result.comments.to_h { |c| [c.location.start_line, c] }
+      visitor = TreeBuilder.new(source, comments_by_line: comments_by_line, file: file)
       visitor.visit(result.value)
       visitor.tree
     rescue ParseError => e
@@ -112,12 +113,13 @@ module RBI
     end
 
     class Visitor < Prism::Visitor
-      #: (String source, file: String) -> void
-      def initialize(source, file:)
+      #: (String source, file: String, ?comments_by_line: Hash[Integer, Prism::Comment]) -> void
+      def initialize(source, file:, comments_by_line: {})
         super()
 
         @source = source
         @file = file
+        @comments_by_line = comments_by_line
       end
 
       private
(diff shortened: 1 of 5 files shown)


lib/rbi/parser.rb
```diff
@@ -183,11 +325,15 @@ class TreeBuilder < Visitor
...
 
-      #: (String source, comments: Array[Prism::Comment], file: String) -> void
-      def initialize(source, comments:, file:)
-        super(source, file: file)
+      #: (String source, file: String, ?comments: Array[Prism::Comment]?,
+      #| ?comments_by_line: Hash[Integer, Prism::Comment]) -> void
+      def initialize(source, file:, comments: nil, comments_by_line: {})
+        if comments
+          comments_by_line = comments.to_h { |comment| [comment.location.start_line, comment] }
+        end
```
KaanOzkan, line 333: Wdyt of raising if both `comments` and `comments_by_line` are supplied so we don't ignore it silently?

If we don't go that route: Is there a downside of prioritizing `comments_by_line` instead of calculating it based on `comments`? Using what's supplied feels better.

lib/rbi/parser.rb
```diff
@@ -991,6 +1076,7 @@ def visit_call_node(node)
 
         visit(node.receiver)
         visit(node.block)
+        @current.comments.concat(comments_inside(node)) if node.message == "sig"
```
KaanOzkan, line 1079: Agent brought up that sig comments aren't printed so this case fails now:

```rb
def test_preserves_comments_inside_sig_in_rbs
  rbi = parse_rbi(<<~RBI)
    sig do
      # keep me
      void
    end
    def foo; end
  RBI

  assert_includes(rbi.rbs_string, "# keep me")
end
```

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+            @comments_by_line.delete(line)
+            next
+          end
+
+          loc = Loc.from_prism(@file, comment.location)
+
+          # If we find the start of a RBS comment, we create a new RBSComment
+          # Note that we ignore RDoc directives such as `:nodoc:`
+          # See https://ruby.github.io/rdoc/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Directives
+          if text.start_with?("#:") && !(text =~ /^#:[a-z_]+:/)
```
amomchilov, line 206: esoteric syntax

```suggestion
          if text.start_with?("#:") && !text.match?(/^#:[a-z_]+:/)
```

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+            text = text.sub(/^#: ?/, "").rstrip
+
+            # If we found continuation comments, we merge them in reverse order (since we go from bottom to top)
+            rbs_continuation.reverse_each do |rbs_comment|
+              continuation_text = rbs_comment.location.slice.sub(/^#\| ?/, "").strip
+              continuation_loc = Loc.from_prism(@file, rbs_comment.location)
+              loc = loc.join(continuation_loc)
+              text = "#{text}#{continuation_text}"
```
amomchilov, line 214: We can avoid the reallocation here


```suggestion
              text.concat(continuation_text)
```

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+            end
+
+            rbs_continuation.clear
+            comments.unshift(RBSComment.new(text, loc: loc))
```
amomchilov, line 218: How about we `<<` them to the end instead, and only `reverse!` at the end?

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+          else
+            # If we have unused continuation comments, we should inject them back to not lose them
```
amomchilov, line 220: Would be helpful to pair these comments with a visual example.

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+            end
+
+            rbs_continuation.clear
+            comments.unshift(parse_comment(comment))
+          end
+
+          @comments_by_line.delete(line)
+        end
+
+        # If we have unused continuation comments, we should inject them back to not lose them
```
amomchilov, line 232: How does this differ from the same logic on line 220?

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+
+        comments
+      end
+
+      #: (Prism::Node node, ?min_line: Integer?) -> [Integer, Integer]
+      def comment_lookup_range(node, min_line: nil)
+        node_start_line = node.location.start_line
+        latest_line = node_start_line
+        comment = @comments_by_line[latest_line]
+        # Start comment lookup from line above the node if there are no trailing comments.
```
amomchilov, line 246: Example?

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+      end
+
+      #: (Prism::Node node, Prism::Comment comment) -> bool
+      def inline_comment_for_node?(node, comment)
+        return false unless node.location.start_line == comment.location.start_line
+
+        between = @source[node.location.end_offset...comment.location.start_offset] || ""
+        between.each_char.all? do |char|
+          char == "," || char.match?(/\s/)
+        end
```
amomchilov, line 268: Is the char at `comment.location.start_offset` the `#`?


```suggestion
        between = @source[node.location.end_offset..comment.location.start_offset] || ""
        between.match?("^,?\s*#$")
```

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+      end
+
+      #: (Prism::Node node) -> Array[Comment]
+      def comments_inside(node)
```
amomchilov, line 272: Doc comment with example please

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+        comments = [] #: Array[Comment]
+
+        node.location.start_line.upto(node.location.end_line) do |line|
+          comment = @comments_by_line[line]
+          next unless comment
+
+          location = comment.location
+          next if location.start_offset < node.location.start_offset
+          next if location.end_offset > node.location.end_offset
```
amomchilov, line 281: What are these checking?

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+        outer_begin_column = outer.begin_column
+        outer_end_line = outer.end_line
+        outer_end_column = outer.end_column
+        return false unless inner_begin_line && inner_begin_column && inner_end_line && inner_end_column
+        return false unless outer_begin_line && outer_begin_column && outer_end_line && outer_end_column
+
+        begins_inside = inner_begin_line > outer_begin_line ||
+          (inner_begin_line == outer_begin_line && inner_begin_column >= outer_begin_column)
+        ends_inside = inner_end_line < outer_end_line ||
+          (inner_end_line == outer_end_line && inner_end_column <= outer_end_column)
```
amomchilov, line 308: Would this be simpler if the calculations were done using the offsets, rather than lines/cols?

Then it just becomes a matter of checking if the `inner` range is a subinterval of `outter` range with `Range#cover?`

lib/rbi/parser.rb
```diff
@@ -174,6 +176,146 @@ def self?(node)
...
+
+        begins_inside && ends_inside
+      end
+
+      #: (Prism::Comment node) -> Comment
+      def parse_comment(node)
+        text = node.location.slice.sub(/^# ?/, "").rstrip
+        loc = Loc.from_prism(@file, node.location)
+        Comment.new(text, loc: loc)
```
amomchilov, line 317: Perhaps this should be a factory function on `Comment`


```suggestion
        Comment.from_prism_node(node)
```

lib/rbi/parser.rb
```diff
@@ -183,11 +325,15 @@ class TreeBuilder < Visitor
       #: Prism::Node?
       attr_reader :last_node
 
-      #: (String source, comments: Array[Prism::Comment], file: String) -> void
-      def initialize(source, comments:, file:)
-        super(source, file: file)
+      #: (String source, file: String, ?comments: Array[Prism::Comment]?,
+      #| ?comments_by_line: Hash[Integer, Prism::Comment]) -> void
+      def initialize(source, file:, comments: nil, comments_by_line: {})
```
amomchilov, line 330: This parameter pack is starting to get unwieldy. I think we should extract a class like something like `SourceFile`, which stores all this stuff

https://refactoring.guru/introduce-parameter-object

lib/rbi/parser.rb
```diff
@@ -608,82 +754,14 @@ def detach_comments_from_sigs(sigs)
...
 
         sigs.each do |sig|
-          comments += sig.comments.dup
-          sig.comments.clear
+          inside, outside = sig.comments.partition { |comment| loc_inside?(comment.loc, sig.loc) }
+          comments.concat(outside)
+          sig.comments.replace(inside)
         end
 
         comments
```
amomchilov, line 762: How do you feel about `flat_map` with a sideffect?

```suggestion
        sigs.flat_map do |sig|
          inside, outside = sig.comments.partition { |comment| loc_inside?(comment.loc, sig.loc) }

          sig.comments.replace(inside)
          
          outside
        end
```

amomchilov: The partition+replace combo is a bit tricky.

What do you think of vendoring a copy of `extract!` from ActiveSupport?

https://github.com/rails/rails/blob/afa0f9244d22abc0b88b46c38c906b518b5fc3c1/activesupport/lib/active_support/core_ext/array/extract.rb#L5-L22

Then this could just be:

```suggestion
        sigs.flat_map do |sig|
          sig.comments.extract! { |comment| !loc_inside?(comment.loc, sig.loc) }
        end
```

lib/rbi/parser.rb
```diff
@@ -783,10 +861,10 @@ def parse_params(node)
 
       #: (Prism::CallNode node) -> Sig
       def parse_sig(node)
-        builder = SigBuilder.new(@source, file: @file)
+        builder = SigBuilder.new(@source, comments_by_line: @comments_by_line, file: @file)
         builder.current.loc = node_loc(node)
         builder.visit_call_node(node)
-        builder.current.comments = node_comments(node)
+        builder.current.comments = node_comments(node) + builder.current.comments
```
amomchilov, line 867: One less alloc + copy


```suggestion
        builder.current.comments = node_comments(node).concat(builder.current.comments)
```

lib/rbi/parser.rb
```diff
@@ -936,11 +1014,16 @@ class SigBuilder < Visitor
       #: Sig
       attr_reader :current
 
-      #: (String content, file: String) -> void
-      def initialize(content, file:)
+      # Bounds sig param comment lookup to comments inside the current `params(...)` call.
```
amomchilov, line 1017: ?

amomchilov (commented):
Great feature!

lib/rbi/parser.rb
Morriar: I want to remove `comments` but I first need to bump Spoom because of the `export-rbi` command.

PR #658, opened 2026-08-28
Do not instantiate `Public` visibility for implicit public defs and attrs

Morriar:
We can use `nil` to represent implicit public default visibility.

cc. @michaelherold.

### DIFF
--- a/lib/rbi/model.rb
+++ b/lib/rbi/model.rb
@@ -326,8 +326,8 @@ module RBI
     #: Array[Symbol]
     attr_reader :names
 
-    #: Visibility
-    attr_accessor :visibility
+    #: Visibility?
+    attr_accessor :visibility # `nil` for implicit `public` visibility
 
     #: -> Array[Sig]
     def sigs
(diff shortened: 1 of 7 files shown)


michaelherold (approved):

lib/rbi/model.rb
```diff
@@ -345,12 +345,12 @@ def sigs?
...
     #|   Symbol name,
     #|   Array[Symbol] names,
-    #|   ?visibility: Visibility,
+    #|   ?visibility: Visibility?,
     #|   ?sigs: Array[Sig]?,
     #|   ?loc: Loc?,
     #|   ?comments: Array[Comment]?
     #| ) -> void
-    def initialize(name, names, visibility: Public::DEFAULT, sigs: nil, loc: nil, comments: nil)
+    def initialize(name, names, visibility: nil, sigs: nil, loc: nil, comments: nil)
```
Morriar: Let's also update `FlattenVisibilities` to keep implicit public as `nil`.

Right now that rewriter seeds/pushes `Public.new` and assigns it to every `Attr`/`Method`, so parsed implicit-public members lose the new representation after `flatten_visibilities!`. That keeps the `Public` allocation/retention on that path.

We can use a `Array[Visibility?]` stack starting with `nil`, push `nil` for new scopes, and only store `Public` when we actually see an explicit `public` label.

lib/rbi/model.rb
```diff
@@ -991,9 +991,6 @@ def initialize(loc: nil, comments: nil, &block)
       super(:public, loc: loc, comments: comments)
       block&.call(self)
     end
-
-    # Shared default instance to avoid allocating a new Public on every Method/Attr creation.
-    DEFAULT = new.freeze #: Public
```
vinistock, line 996: If the implicit one is now represented by `nil`, could we make `loc` required for the visibility instances?

vinistock (approved):

